PR #4, opened 2020-05-14
Create LICENSE

dbenn:
Adding the AGPL licence deed to GitHub. We have always used the Affero GPL 3 licence for VStar (see doc/license).

Please confirm you're happy with this.


SaraAAVSO:
If you are happy with this David, then so am I.

dbenn:
Done.

Issue #12, opened 2020-05-17
Common VELA filter does not work in KeplerFITSObservationSource


dbenn:
### Common VELA filter does not work in KeplerFITSObservationSource 

**reason**: missing constructor of class KeplerFITSObservationRetriever
**should be**:
~~~
public KeplerFITSObservationRetriever {
    super(getVelaFilterStr());
}
~~~
From https://sourceforge.net/p/vstar/bugs-and-features/708/ by @mpyat2

mpyat2:
@dbenn: already fixed, can be closed

PR #16, opened 2020-05-30
#12: added constructor which passes VeLa filter string to base class

dbenn:
Added constructor as per Max's proposal. Appears to work correctly now.

### DIFF
--- a/plugin/src/org/aavso/tools/vstar/external/plugin/KeplerFITSObservationSource.java
+++ b/plugin/src/org/aavso/tools/vstar/external/plugin/KeplerFITSObservationSource.java
@@ -17,42 +17,37 @@
  */
 package org.aavso.tools.vstar.external.plugin;
 
+import java.awt.Color;
+import java.awt.Container;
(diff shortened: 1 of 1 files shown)


mpyat2:
OK, I see changes. I compiled the plugin and tested: now VeLa filter works.

plugin/src/org/aavso/tools/vstar/external/plugin/KeplerFITSObservationSource.java
```diff
@@ -171,6 +166,10 @@ public AbstractObservationRetriever getObservationRetriever() {
 		private BasicHDU[] hdus = null;
 		
 		private String objName = null;
+		
+		public KeplerFITSObservationRetriever() {
+			super(getVelaFilterStr());
```
mpyat2, line 171: OK

dbenn:
> I see the mechanics, but don't really understand the code change.

No worries Cliff. We're just passing the VeLa filter string from the plugin to the observation retriever.

PR #24, opened 2020-06-07
#23: added Unicode support to VeLa identifiers; added a unit test for…

dbenn:
… this; added Unicode symbol as a synonym for PI to VeLa interpreter; incodentally also added beginning of VeLa based standard functions file and unit tests; VeLa test formatting changes

### DIFF
--- a/script/VeLa/fact.vl
+++ b/script/VeLa/fact.vl
@@ -1,11 +1,11 @@
 #!/usr/bin/env /Users/david/vstar/VeLa.sh --verbose
 
-fact(n:integer) : integer {
+n!(n:integer) : integer {
     when
         n <= 0 -> 1
-        true -> n*fact(n-1)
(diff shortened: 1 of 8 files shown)


dbenn:
@clkotnik and @mpyat2: What I'm doing here is to allow Unicode characters in VeLa identifiers. What this allows is for identifiers like π instead of PI. I may eventually add Unicode to strings but identifiers are something I've wanted to add Unicode support to for awhile. Don't stress too much about this but feel free to ask questions and indicate if you are OK for me to merge it.

mpyat2:
OK, so we can use greek letters, for example. And even use something
like чудернацький_ідентифікатор, ha-ha!

dbenn:
> чудернацький_ідентифікатор

Yep, "whimsical_identifier" :)

PR #29, opened 2020-06-10
VSX Query plig-in: initial upload.

mpyat2:
New plug-in VSX Query. See documentation VSX Query Plug-in.pdf


dbenn:
Cool. Thanks Max!

Issue #7, opened 2020-05-17
Plugin Management Fix (Deletion under Windows)


dbenn:
Now all plugin operations work under Windows (i.e. Delete)

src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManagementDialog.java
src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
src/org/aavso/tools/vstar/ui/dialog/prefs/PluginSettingsPane.java
src/org/aavso/tools/vstar/ui/MenuBar.java
src/org/aavso/tools/vstar/ui/resources/PluginLoader.java

1. A list for URLClassLoader-s for all plugin classes
2. Closing URLClassLoader list before delete/update operations (PluginManagementDialog, PluginSettingsPane)
3. All temporary streams and ClassLoaders created in PluginManager get closed
4. Checking result of delete operations PluginManager
5. A button encouraging a user to close (and restart) VStar after plugin operations (PluginManagementDialog) and a message to a user after deleting all plugins (PluginSettingsPane)
6. I tried to prevent double-starting PluginManagementDialog from the menu. See MenuBar, PluginManagementDialog, PluginSettingsPane

dbenn:
From:
* https://sourceforge.net/p/vstar/bugs-and-features/625/
* https://sourceforge.net/p/vstar/bugs-and-features/707

dbenn:
Duplicate of #33

PR #33, opened 2020-06-17
Attempt to fix plugin deletion error under Windows.

mpyat2:
Issue #7: plug-ins cannot be deleted under Windows.
The fix attempts to resolve the issue: ClassLoader-s and streams are closed while plugin management.
Also, a small piece of code has been added to MenuBar.java to prevent starting PluginManagementDialog twice (currently this is possible even if PluginManagementDialog were modal: there is enough time to click the menu many times while initialization of PluginManagementDialog).
After any plugin-management-related operation, a user prompted to restart VStar.

### DIFF
--- a/src/org/aavso/tools/vstar/ui/MenuBar.java
+++ b/src/org/aavso/tools/vstar/ui/MenuBar.java
@@ -1199,6 +1199,10 @@ public class MenuBar extends JMenuBar {
 		return new ActionListener() {
 			@Override
 			public void actionPerformed(ActionEvent e) {
+				// Max: probably I'm paranoid: preventing opening PluginManagementDialog twice.
+				if (PluginManagementDialog.getPluginManagementActive()) return;
+				PluginManagementDialog.setPluginManagementActive(true);
+				//System.out.println(Thread.currentThread().getId());
 				final PluginManager manager = new PluginManager();
 				PluginManagementOperation op = new PluginManagementOperation(
 						manager, "Initialising Plug-in Manager") {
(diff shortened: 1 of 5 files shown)


src/org/aavso/tools/vstar/ui/MenuBar.java
```diff
@@ -1199,18 +1199,40 @@ public ActionListener createPluginManagerListener() {
 		return new ActionListener() {
 			@Override
 			public void actionPerformed(ActionEvent e) {
+				// Max: probably I'm paranoid: preventing opening PluginManagementDialog twice.
```
dbenn, line 1202: A reasonable concern in fact. Thanks, well spotted.

src/org/aavso/tools/vstar/ui/MenuBar.java
```diff
@@ -1199,18 +1199,40 @@ public ActionListener createPluginManagerListener() {
...
+				if (PluginManagementDialog.getPluginManagementActive()) return;
+				PluginManagementDialog.setPluginManagementActive(true);
```
dbenn, line 1204: It's probably best to set this after line 1233 when the plugin manager has been fully activated. Indeed, one possible scenario is that the user clicks multiple times, causing this method to be invoked more than once, such that execution of a thread may be suspended after the active setter on line 1204 while a second enters the method, going no further than 1203 because the flag has already been set. The first thread would still resume to execute the rest and invoke the manager though, so probably okay. Adding the synchronised keyword to the method would allow the whole method to be treated as a critical section. That should work well with the active flag setting at line 1233. Am I making sense? It's later here. :)

dbenn:
I have no easy access to a Windows laptop at home. I'm guessing you've tested this on Windows @mpyat2

src/org/aavso/tools/vstar/ui/MenuBar.java
mpyat2: I think no. This code is intended to prevent the secondary execution of the menu action so it should be at the top of the "actionPerformed". The same check is done in PluginSettingsPane.java, the second place where one of the plugin management operations (deleteAllPlugins) is executed. "pluginManagementActive" flag can be placed elsewhere, PluginManagementDialog is not the unique place. pluginManagementActive flag is normally accessed from the main thread except one place: an exception handler at the line 1221 of MenuBar.jar where the flag is reset. This is why I declared it as "volatile" initially, yet synchronized getter and setter probably even better.

clkotnik:
I built this branch on my Windows 10 PC and tested it.  I deleted individual plugins, added individual plugins and exercised the delete all plugins from the prefs dialog.  In all cases it behaved as expected.

mpyat2:
Hi Cliff,

Thank you!

Cheers,
Max

dbenn:
@mpyat2 @clkotnik Hi Max and Cliff. I've carried out those tests on the branch under Mac OS X to make sure it still works. All good!

Thanks guys. 

Max, these changes will be appreciated by Brad's course participants in August, and all Windows users. Appreciate it!

I'll merge the branch now.

Issue #19, opened 2020-06-01
setWindowAlwaysOnTop permission error when HarmonicInputDialog invoked


dbenn:
A CHOICE course participant found another occurrence of the `setWindowAlwaysOnTop` permission error in a WebStart context, this time in the HarmonicInputDialog. Here's the logged error:
~~~
Jun 01, 2020 12:17:59 AM org.aavso.tools.vstar.ui.dialog.MessageBox showErrorDialog
SEVERE: Modelling Error
java.security.AccessControlException: access denied ("java.awt.AWTPermission" "setWindowAlwaysOnTop")
	at java.security.AccessControlContext.checkPermission(Unknown Source)
	at java.security.AccessController.checkPermission(Unknown Source)
	at java.lang.SecurityManager.checkPermission(Unknown Source)
	at com.sun.javaws.security.JavaWebStartSecurity.checkPermission(Unknown Source)
	at java.awt.Window.setAlwaysOnTop(Unknown Source)
	at org.aavso.tools.vstar.ui.dialog.model.HarmonicInputDialog.<init>(Unknown Source)
	at org.aavso.tools.vstar.external.plugin.FourierModelCreator.getModel(Unknown Source)
	at org.aavso.tools.vstar.ui.mediator.Mediator.performModellingOperation(Unknown Source)
~~~

dbenn:
This is similar to:

- https://sourceforge.net/p/vstar/bugs-and-features/701/
- https://sourceforge.net/p/vstar/bugs-and-features/628/

dbenn:
This error is also happening with the Fourier model plugin according to a CHOICE course participant.

dbenn:
So too for Save dialogs, at least on the Mac.

mpyat2:
After removing this.setAlwaysOnTop(true); the dialog works normally (Windows 7)

PR #43, opened 2020-08-18
#19: removed window set-on-top call

dbenn:


clkotnik:
I think this is in the JNLP test jar that Sara recently placed on a test site.  I verified that this test version handle the fourier model without the security error on Win 10 with java version "1.8.0_261".

If I am confusing this with another issue, please pardon me.

PR #60, opened 2020-09-17
"Magnitude median" added.

mpyat2:
I've added "magnitude median" to "Descriptive statistics by series" plugin.
Sometimes, while estimating the base magnitude level of the different datasets, the median works better than the mean, especially for eclipsing stars.


dbenn (approved):
Looks good @mpyat2

PR #63, opened 2020-10-01
SuperWASP TAMFLUX2 support

mpyat2:
There are two versions of data in SuperWASP FITS file: FLUX2 and TAMFLUX2 (https://exoplanetarchive.ipac.caltech.edu/docs/SuperWASPProcessing.html). The situation is similar to TESS/Kepler: the first set is ‘raw’, the second one is ‘corrected’ (TAMFLUX2 = ‘SYSREM corrected fluxes’).

Existing SuperWASP plugin reads FLUX2 (‘raw’). The quality of the ‘raw’ data is too bad. On the other hand, magnitudes calculated from TAMFLUX2 (corrected fluxes) are the same as in CSV files that can be downloaded along with FITS from the SuperWASP portal (https://wasp.cerit-sc.cz/form).

I’ve been using SuperWASP CSV data for a long time and noticed quite long ago that they differ from those derived by VStar plugin from FITS. However, only now I investigated SuperWASP FITS and found the source of the discrepancy.

I’ve modified the SuperWASP plugin so that a user can choose a version of data. By default, the corrected (TAMFLUX2) version is selected.

I've also made the SuperWASP plugin 'heliocentric'

### DIFF
--- a/plugin/src/org/aavso/tools/vstar/external/plugin/SuperWASPFITSObservationSource.java
+++ b/plugin/src/org/aavso/tools/vstar/external/plugin/SuperWASPFITSObservationSource.java
@@ -19,17 +19,14 @@ package org.aavso.tools.vstar.external.plugin;
 
 import java.awt.Color;
 import java.awt.Container;
-import java.awt.Dimension;
-import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
 import java.util.ArrayList;
 import java.util.List;
 
 import javax.swing.BorderFactory;
 import javax.swing.BoxLayout;
-import javax.swing.JButton;
-import javax.swing.JDialog;
+import javax.swing.ButtonGroup;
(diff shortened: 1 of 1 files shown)


clkotnik:
These sound like good changes.  I'm just curious, is there an advantage to using the FITS corrected to just using the CSV?

mpyat2:
SuperWASP CSV file cannot be loaded directly.
Minimal changes required to make it loadable via Flexible Text Format plug-in is to add the following header:
`#FIELDS=TIME,,MAG,MAGERR`
(there is a camera id between the time and magnitude column)

dbenn (commented):
Looks good @mpyat2. Thanks. There are a bunch of print statements starting around line 172. I guess you are planning to remove them but just mentioning it in case you forgot.

dbenn:
Thanks @mpyat2 !

Issue #52, opened 2020-08-29
Loading saved residuals fails without deleting extra empty columns


dbenn:
Reported by two data analysis with VStar CHOICE course participants (Ken and Sean).

See also https://www.aavso.org/comment/72318#comment-72318

dbenn:
Note that I have not seen this before.

dbenn:
Given that file saving is now an observation sink plugin (this type was introduced in 2.21.1), the problem may lie there.

dbenn:
The bug was introduced here: https://github.com/AAVSO/VStar/commit/cc6adcc5f5f03f76f3a4c6bb5f31f4f11dde8711#diff-871f4a5a2a555153a3fea1ec8a87894bR2380 (see line 2380). Given that the save mechanism was replaced in 2.21.2 by a generic observation sink plugin mechanism, the simplest approach is to create separate plugins for download and simple formats. This distinction was previously made by the Mediator caller code by a parameter to the ObservationSaveTask constructor.

PR #64, opened 2020-10-04
#52: implemented two specific obs sink plugins for simple and download types

dbenn:
Hi @mdadighat. When you get the chance, I'd appreciate your review of the changes I've made for this save residuals bug. Thanks.


dbenn:
Also, @mdadighat let me know if you have questions about what I've done.

https://github.com/AAVSO/VStar/pull/64/files

src/org/aavso/tools/vstar/plugin/ob/sink/impl/TextFormatObservationSinkPlugin.java
```diff
@@ -1,142 +0,0 @@
...
-		SUFFIXES.put("Comma", "csv");
-		SUFFIXES.put("Space", "txt");
-		SUFFIXES.put("Tab", "tsv");
-	}
-
-	@Override
-	public void save(PrintWriter writer, List<ValidObservation> obs,
-			String delimiter) throws IOException {
-
-		if (Mediator.getInstance().getLatestNewStarMessage().getNewStarType() == NewStarType.NEW_STAR_FROM_SIMPLE_FILE) {
```
mdadighat, line 61: Was the issue that this check wasn't behaving as expected when saving residuals? I.e. it thought it should be saved in simple format when it shouldn't since it wasn't a new star?

mdadighat (approved):

src/org/aavso/tools/vstar/plugin/ob/sink/impl/TextFormatObservationSinkPlugin.java
dbenn: Hi @mdadighat. Yes, that's pretty much it.

Figuring out why what's changed, has changed, based upon the diffs is probably quite confusing here.

This check on line 61 above, that you show, of the initial implementation of the text format observation sink plugin made an incorrect assumption about what format to save the file as, based upon the latest dataset loaded.

I honestly look at that initial plugin code now and I wonder what I was thinking.

Before the addition of the observation sink plugin interface, the Mediator code passed the so-called "new star type" (simple format, AAVSO download format, arbitrary format) as a parameter to the ObservationSaveTask. After the plugin interface was added, the plugin tried to determine this "new star type" on its own, incorrectly.

The plugin initially appeared as a unified "Simple or Download" format item in the save dialog, whereas now, there are two plugins, for simple and download. Now there is no ambiguity. Either we are saving as simple format or download format, explicitly.

I hope that makes sense.

mdadighat: Yep, makes sense! Splitting the plugin into two seemed straight forward but I just wanted to double check that I understood the specific reason for it. I wonder if this was the only case affected or just the only one that got noticed.

dbenn: This was the only case because the only two save mechanisms before the obs sink plugin mechanism was added.

Issue #67, opened 2020-10-24
HJD converter declination dialog does not handle negative degrees


dbenn:
Robert Jenkins reported not being able to enter negative degrees when into the HJD converter declination dialog.

I've just confirmed this. I'm surprised not to have seen this before now.

PR #69, opened 2020-10-24
#67: fixed dec min/max in dialog field; also fixed RA min/max

dbenn:
In both cases, max was null and min was 0 or previous value.

@clkotnik, @mdadighat, @mdadighat: if anyone has the chance to review this change, that'd be great. Thanks.

### DIFF
--- a/src/org/aavso/tools/vstar/ui/mediator/Mediator.java
+++ b/src/org/aavso/tools/vstar/ui/mediator/Mediator.java
@@ -1479,9 +1479,9 @@ public class Mediator {
 			s = hms.third;
 		}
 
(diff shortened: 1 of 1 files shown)


mdadighat (approved):

mdadighat:
I'm assuming null wasn't ever an appropriate value for the min, max, or initial value to be in these cases?

dbenn:
> I'm assuming null wasn't ever an appropriate value for the min, max, or initial value to be in these cases?

The IntegerField (and DoubleField) uses null to mean no lower or upper valid bound. In more recent times (e.g. in VeLa) I've started making use of Optional values.

PR #71, opened 2020-10-24
Issue 70 double field set value issue

mpyat2:
DoubleField.setValue() is locale-specific now and is consistent with getValue()


mpyat2:
DoubleField.setValue() fix.

dbenn (approved):
Looks good

dbenn (commented):
Thanks Max

Issue #62, opened 2020-09-26
Harmonic search needs to use a numeric tolerance vs equality approach


dbenn:
See findHarmonics() method in PeriodAnalysisDialogBase.

dbenn:
@mdadighat I've been meaning to give an example of a method I've used in some of the unit test code for determining whether two values are close with respect to some tolerance (epsilon) value:
~~~
	private boolean areClose(double a, double b, double epsilon) {
		return Math.abs(a - b) < epsilon;
	}
~~~

mdadighat:
@dbenn I originally was thinking something like that but then I went down a rabbit hole of attempting to account for significant figures and rounding and probably overcomplicated it. I had some issues with really small values (frequencies of .000711 and similar) but I don't know if that's part of the issue this is trying to solve.

dbenn:
Hi @mdadighat I do completely understand about rabbit holes. I got down them on a regular basis. It's okay. In my experience, no solution is perfect. :) The question I would ask myself is whether some solution is better than what exists.

PR #72, opened 2020-10-25
#62: Use tolerance range instead of exact match with findHarmonics

mdadighat:
In testing how narrow of a range to use - 2 decimal places added way too many more harmonics that didn't seem to have much value and 4 or 5 decimal places didn't have much of an effect since it was close to how far apart the harmonic values were anyway. Due to that I didn't see much of a place for there to be a preferences value for this tolerance but I don't know if I could be missing a case.


dbenn:
Hi @mdadighat What you say seems reasonable to me. Your current 0.001 tolerance is probably fine in practice. The only thing I wonder about is whether the desirable/useful tolerance differs across variable types. @clkotnik, @mpyat2: any thoughts?

dbenn (commented):
This 0.001 tolerance looks good to me @mdadighat. The only thing I wonder about is whether the useful tolerance could vary for variable type. @clkotnik or @mpyat2: any thoughts on this? Otherwise, we should just go with this change because it will be better than what we have now.

mpyat2:
Probably it is Ok for most cases. However, there are variables having short periods (~0.1 day or less, Delta Scuti variables, for example). Could the tolerance be relative, not absolute?

dbenn:
@mpyat2 That's the kind of thing I was wondering about. Are you thinking of something like the relative tolerance of Python functions such as numpy.isclose(), math.isclose(), https://docs.pytest.org/en/latest/reference.html#pytest-approx?

There are probably R equivalents that are relative tolerance based. This one I think is absolute: https://stat.ethz.ch/R-manual/R-devel/library/base/html/all.equal.html

I've been using C++ BOOST UT lately. It has BOOST_CHECK_CLOSE (https://www.boost.org/doc/libs/1_35_0/libs/test/doc/components/test_tools/reference/BOOST_CHECK_CLOSE.html). That's percentage based but still absolute as I recall.

Thoughts @mdadighat ?

mdadighat:
In cases where the period is shorter would you want a smaller tolerance or larger one? I checked a few different variables and found the frequencies were of a similar precision so I wasn't sure how the star's period would affect things here - i.e. how the tolerance should relate. Although a relative tolerance might be better anyway, I just wanted to confirm how it would be expected to affect different cases so I could test it. I'll look into options on making it relative.

mdadighat:
@dbenn @mpyat2 I didn't see a built in relative tolerance and didn't want to import any additional libraries so I looked at the implementation of the above examples and ended up with using the "strong" method (check should be true when comparing the relative tolerance value for both **a** and **b** values as opposed to true for just one of them, which was the "weak" method) of the Python implementation of isClose. 

It was pretty straightforward and testing it with a wider range of values seemed to get behavior that is more like what you were describing. My only other question would be if this belongs in a library somewhere in the code so it could be reused, and if so where would be the most appropriate place for it?

dbenn:
Apologies for the delay. This is great! Thanks @mdadighat! I agree that a diff/tolerance utility class of some kind would be awesome. Perhaps under org.aavso.tools.vstar.util?

mdadighat:
@dbenn I moved it over and also moved the checks you referenced from the tests there as well so it can be a single absolute/relative method depending on what's needed.

dbenn:
Looks good @mdadighat. Thanks! On line 43 of the Tolerance class, if you wanted to avoid the need for explicitly returning true or false, that could become:
~~~
return ( (diff == 0.0) || ...
~~~
instead of:
~~~
if( (diff == 0.0) || ...
~~~
Either is fine. Just a thought.

Looking over a number of the VeLaTest cases where I used 1e6, I'm wondering whether I meant to use 1e-6. It's late though, so I'll check tomorrow.

dbenn:
@mdadighat I've fixed those UT tolerances I mentioned above. Happy to merge if you are.

dbenn (approved):

mdadighat:
@dbenn When I went back it did look clunkier than I thought so I changed it. Do you usually merge these or should I (after approval)?

dbenn:
@mdadighat A lot of my code looks clunkier that I remember when I go back to it. :) Looks good. I'm doing the merges at the moment, but that may change down the track.

Issue #97, opened 2020-12-22
Check binary table field names to be sure that correct FITS is used


mpyat2:
Plugins that use FITS format: check binary table field names to be sure that correct FITS is used.

PR #98, opened 2020-12-24
FITS plugins(KeplerFITSObservationSource, LightKurveFITSObservationSo…

mpyat2:
…urce, SuperWASPFITSObservationSource) check for binary table field names to be sure that the correct FITS is used.

Currently, we have 3 plugins that read FITS of different formats. Plug-ins reference data in the FITS binary table by column numbers (different for different formats). To avoid incorrect read, I inserted a piece of code in each plugin that checks for field names.
Issue #97


clkotnik (approved):
I think we can expect these formats to change over time so it is good to have this as a sanity check.

dbenn (approved):
Looks good. Thanks @mpyat2.

Issue #112, opened 2021-01-19
Javadocs do not open from downloaded zip, DMG


dbenn:
@yorkpf reported this today. It used to work. Odd.

dbenn:
@yorkpf reported this today. It used to work. Odd.

dbenn:
Download any of the win, mac or bash archives from https://github.com/AAVSO/VStar/releases/tag/2.21.2 

Navigate to `doc/vstar_docs` in the uncompressed VStar archive (under `Extras` folder in Mac DMG, `vstar` dir in bash and win) and open `index.html` in a browser. Selecting any package or class does nothing at all!

This was not always the case and certainly not last time I looked. @yorkpf and I have seen this on the Mac. Can someone else reproduced this on Windows or Linux? (@mdadighat, @mpyat2, @clkotnik, @SaraAAVSO)

mpyat2:
I tried vstar-bash-2.21.2.zip, vstar-win-2.21.2.zip (under Windows)
Yes, I cannot see any doc (see screenshot below)
This is because 'vstar/doc/vstar_docs/org/' directory is EMPTY (in ZIP and
in unzipped bunch)

[image: image.png]

dbenn:
Ah, yes. Quite right! Sorry.

I've just tracked this down. It's an error in the Ant build targets. Need `<include name="vstar_docs/*" />` to become `<include name="vstar_docs/**" />`. Will sort this out.

Thanks!

mpyat2:
By the way, I realized that vstar_docs is completely missing in the Windows
EXE installer ( VStarWinSetup_2.21.2.zip
<https://github.com/AAVSO/VStar/releases/download/2.21.2/VStarWinSetup_2.21.2.zip>).
This is because I've never run 'javadoc' build option.
Now I tried it and rebuilt vstar-win.zip again:
ant javadoc
ant win

'ant javadoc' generates normal vstar_docs directory, all documents are
accessible.
However, 'ant win' makes vstar-win.zip with empty
vstar/doc/vstar_docs/org/.

BTW, Javadoc displays a lot of errors while building documentation.

Cheers,
Max

mpyat2:
Indeed, with this fix, all documents included in the zip!

dbenn:
Yes, ant win, ant bash, ant mac all suffer from the same problem. Thanks for the confirmation. Will fix this and seek review.

dbenn:
[...]

PR #113, opened 2021-01-24
#112: fixed javadoc file copy instances

dbenn:
Fixed the file copying elements in Ant build file.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -159,7 +159,7 @@
 				<include name="license/*" />
 			</zipfileset>
 			<zipfileset dir="${doc}" prefix="vstar/doc">
-				<include name="vstar_docs/*" />
+				<include name="vstar_docs/**" />
 			</zipfileset>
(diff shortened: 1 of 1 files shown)


yorkpf:
David: Downloaded zip file seems to work as it should. Thanks. For my part, can I safely just replace my current (non-working) vstar_docs with this newer version? Paul

yorkpf (commented):
OK. I can read the new downloaded API.

mpyat2:
Win distribution: created successfully, both vstar-<version>-win.zip and VStarSetup_<version>.exe.
As I see, it is required to run 'ant javadoc' before creating a distribution package (i.e. ant win).
What do you think about making "win", "bash", "mac" targets dependent on "javadoc" target?

dbenn:
When you're happy with the dependent targets @mpyat2, I'll merge this one.

mpyat2:
@dbenn Yes all is OK! Thank you!

Issue #58, opened 2020-09-12
VeLa variables may undesirably change type


dbenn:
This is unintended.

dbenn:
This for example should cause VeLa to stop in its tracks:
```
x <- 2
x <- "2!"
```

dbenn:
Added a unit test along the lines above. It fails as expected with the current implementation.

dbenn:
Try entering code such as:
~~~
x <- 2
x <- "2"
~~~
to see what is permitted and what is not. The above leads to this error:
~~~
The type of the new value ("2") is not compatible with the bound type of X.
~~~
The key point is that unless a value is assigned to a variable whose initial value's type is compatible with it, an error will result. This is OK for example:
~~~
x <- 2.5
x <- 2
~~~
Reversing those lines will yield an error however, since the first value would be an integer whereas the second would have type real, a "downgrade" in the sense of precision. I'm trying to be careful about permitting

PR #118, opened 2021-02-14
#58: VeLa variables may undesirably change type

dbenn:
Fixed by converting new value's type to the type of the existing binding or throwing an exception; added comments and deprecated a function.

### DIFF
--- a/src/org/aavso/tools/vstar/vela/VeLaEnvironment.java
+++ b/src/org/aavso/tools/vstar/vela/VeLaEnvironment.java
@@ -67,6 +67,7 @@ public class VeLaEnvironment<T> {
 	 * @param name
 	 *            The name of the binding to lookup.
 	 * @return Whether the name is bound in this environment.
+	 * @deprecated less useful now, given the use of Optional values
 	 */
 	public boolean hasBinding(String name) {
 		return cache.containsKey(name);
--- a/src/org/aavso/tools/vstar/vela/VeLaInterpreter.java
+++ b/src/org/aavso/tools/vstar/vela/VeLaInterpreter.java
@@ -935,7 +935,7 @@ public class VeLaInterpreter {
 					bound = true;
 				} else {
 					throw new VeLaEvalError(
-							String.format("The type of the new value (%s) is not compatible with the bound type of %s.",
+							String.format("The type of the value (%s) is not compatible with the bound type of %s.",
 									value, name));
 				}
 				break;
--- a/src/org/aavso/tools/vstar/vela/VeLaInterpreter.java
+++ b/src/org/aavso/tools/vstar/vela/VeLaInterpreter.java
@@ -876,7 +876,9 @@ public class VeLaInterpreter {
 	/**
 	 * Unify operand types by converting both operands to strings if only one is a
 	 * string or both operands to double if only one is an integer. We change
-	 * nothing if either type is composite or Boolean.
+	 * nothing if either type is composite or Boolean. Note that this method
+	 * is intended to be used for expressions, not for variable bindings where
+	 * much less conversion/coercion is possible. 
 	 * 
 	 * @param a The first operand.
 	 * @param b The second operand.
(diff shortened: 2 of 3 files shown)


yorkpf:
David

I have started to look at this.

I guess that “reviewing” involves looking at #58 (to understand the nature of the issue) and also looking at #118 (to understand the proposed solution).

The 3 links below, marked with an “M” appear to be the file changes (+ and -) proposed, but point to different portions of the same document that comes up when you click on “Files Changed” on the github site. I understand I need to look at these but what about the two links below them (one ends in .patch and one in .diff) - what do I do with these?

Paul

dbenn:
That’s all correct. You don’t need to worry about the diff and patch files. They can be used with particular tools to modify source code.

David

dbenn:
@mpyat2 any comments on this one?

mpyat2:
@dbenn no comment at this time.

yorkpf:
David

A trivial question: How do you pronounce" Vela” - to rhyme with “wheeler” or “weller”?

More importantly, what was the motivation for developing VeLa? I have read the material at github.com/AAVSO/VStar/wiki/VeLa <http://github.com/AAVSO/VStar/wiki/VeLa>, but I am far from clear that a Java construct of some kind could not achieve the same thing.

Regards
Paul

dbenn:
Paul

To rhyme with wheeler. As I thought about the acronym, it started to look astronomical after the Vela pulsar, SN remnant, and constellation.

The motivation was to create first an expression language that could be used where numbers were used in dialogs (e.g. create a phase plot or model with a period multiplied by 2 or halved, subtract ~10 years from a JD in AID load, e.g. 2458535.5-100*365.25), then to create filters, and then the other items in the list on that page emerged. Not on that list is: scripting VStar itself as an alternative to JavaScript.

But that is not the question you are asking, right? Why create a new language?

I needed something tightly integrated with VStar. I could have made that language Java and called the Java compiler from within VStar (easy enough), but that language is not suitable for simple expressions or scripting nor would many people wish to use it. I had already added JavaScript as a scripting language through Tool -> Run Script... but the JavaScript implementation available in Java changed a couple of times and I didn't want to be at the whim of Sun/Oracle. JavaScript is also too dynamically typed for my taste and I wanted to exclude as many errors as possible in this context. I could have chosen some other scripting engine language such as JRuby or Python but this engine is intended more for scripting something from the outside. As it turns out, I can use VeLa for that too. There were probably other options but I decided in the end to create something (a domain specific language) that would integrate with VStar in the way I wanted, allow localised numbers etc. I had the experience to do that. 

Why a new language, rather than an implementation for a suitably well typed existing language? The answer is less clear and more along the lines of: because I wanted to and I could. 

Maybe I should add some of this detail to the wiki page?

I'd like to hear more about your thought that "...a Java construct of some kind could not achieve the same thing."

Thanks for questioning my approach. It's fine.

David

yorkpf:
Thanks for that reply, David.

I can certainly understand “because I wanted to and I could”. Yes, I can definitely relate to that :-)

The explanation you gave is the gist of what I was looking for:  "that language [Java] is not suitable for simple expressions or scripting nor would many people wish to use it. I had already added JavaScript as a scripting language through Tool -> Run Script... but the JavaScript implementation available in Java changed a couple of times and I didn't want to be at the whim of Sun/Oracle. JavaScript is also too dynamically typed for my taste and I wanted to exclude as many errors as possible in this context. I could have chosen some other scripting engine language such as JRuby or Python but this engine is intended more for scripting something from the outside".

Re my comments (below) to the effect that plain Java ought to be able to achieve the same thing, that was merely speculative - I was just thinking out loud. My knowledge of Java is not deep enough to even hazard an opinion on that one. I was just curious as to your intentions.

BTW, I suppose you know that “vela” in Latin means “sails”, ie. the sails of a sailing ship?

Thanks.
Paul

dbenn:
Thanks Paul.

I was aware of the sails/ship connection but had forgotten. VeLa puts the wind in your sails? ;) Vela, Carina, Crux, Centaurus together make up a very rich and beautiful part of the sky.

I appreciate my perspective and motives being challenged. Keeps me on my toes.

David

dbenn:
Added a brief excerpt of the foregoing to the wiki Intro section.

yorkpf:
Wasn’t so much a challenge, just an enquiry :-)

Issue #121, opened 2021-02-19
Chart Properties tab in Preferences


mpyat2:
Sometimes visible gridlines are useful. Currently, the user does not have access to gridlines color, they are white and invisible. An awkward way to make them visible is changing the background color via a chart popup dialog. This change is not persistent.

To solve this, a new "Chart Properties" tab is introduced, see a new branch "chart-general-preferences" :
![image](https://user-images.githubusercontent.com/44919945/108509663-2f71a480-72c6-11eb-97ce-dba441f426b1.png)

Other chart properties can be added to this tab.

mpyat2:
This is not a 100% ideal solution. Background color still can be changed by chart popup dialog and became not synchronized with Properties tab. Such a change is temporary, the color of newly created plots will be taken from Chart Properties. I think that such a little inconsistency can be neglected.

dbenn:
I think it's a good solution. One thing that occurred to me after merging is that using localisation in main dialogs and UI items is good where possible, but it's not used everywhere and going through everything is a larger task. So, no need for us to worry about this at the moment.

PR #122, opened 2021-02-19
"Chart Properties" tab in Preferences.

mpyat2:
See issue #121


dbenn (approved):
Yes, this is a good change and something I've gone back and forth on. I like it. Thanks!

PR #126, opened 2021-03-03
Documentation updated

yorkpf:


dbenn (changes requested):
Hi @yorkpf. I'm seeing an error when I go to:

http://cdsarc.u-strasbg.fr/viz-bin/nph-Plot/Vgraph/htm?1/239/834

Looks like there's a "?1" instead of a "?I" (capital "i"). So if I use this, it works:

http://cdsarc.u-strasbg.fr/viz-bin/nph-Plot/Vgraph/htm?I/239/834

yorkpf:
Thanks David. I thought I had fixed that. How do I fix that fix? Just start me off ...

Paul

dbenn:
Hi Paul. One way would be to go to:

https://github.com/AAVSO/VStar/blob/Issue-119-Update-documentation-in-HipparcosObservationSource-Plugin/plugin/src/org/aavso/tools/vstar/external/plugin/HipparcosObservationSource.java

and click the pencil icon to edit it and make the change there. The alternative is to make it in Eclipse then commit and push, but in this case, editing in place in the web page makes good sense.

yorkpf:
OK. Done ...

dbenn (approved):
That URL works fine. Thanks.

Issue #127, opened 2021-03-05
VeLa dialog does not show any output


mpyat2:
![image](https://user-images.githubusercontent.com/44919945/110138598-cf592300-7dda-11eb-8615-9affa575d2f4.png)

No output after [RUN].

dbenn:
Hi @mpyat2, yes it's true. I have relied upon the use of print() here. Given that the main use for that is really in the VeLa dialog, I agree that at least the last expression should be output. If other intermediate expressions are desired to be output, print can still be used. I initially thought of this as like a REPL, but then shied away from that. No very good reason though. I must have made that change between first writing https://github.com/AAVSO/VStar/wiki/VeLa because the dialog under `Running VeLa Code` there shows output. Perhaps that's what led you to this bug? Anyway, I agree with fixing it.

dbenn:
Perhaps the wiki page needs to be updated too (in the section I mentioned above)?

mpyat2:
Thank you, David. Indeed,  I read the old Sourceforge Wiki and tried the example.

PR #128, opened 2021-03-05
VeLa dialog "no output" fix.

mpyat2:
Issue #127: Fixed


dbenn (approved):
Indeed, despite what I said in #128, I now see that this was simply an oversight on my part when I made a change to the output code at some point. Thanks for the fix.

Issue #117, opened 2021-02-13
JFreeChart library update


mpyat2:
VStar uses JFreeChart version 1.0.13, dated 20 Apr 2009. An update to the most recent version 1.5.2 (31 Dec 2020) is proposed.

dbenn:
I notice that 2.0.0 is in the works (https://github.com/jfree/jfreechart) but there may be more compatibility issues here. I note that pseudo 3D plots have gone away and they defer to Orson Plots for 3D (http://www.object-refinery.com/orsoncharts/).

mpyat2:
I compiled jfreechart-1.6.0-SNAPSHOT.jar (from code https://github.com/jfree/jfreechart) using maven. It requires JDK 11 to compile. I see API changes: getXYPlot method does not exist anymore. So the transition to JFreeChart >1.5 is problematic.

dbenn:
Looks like something to keep an eye on for the future then eh? 1.5 seems like the way to go. Perhaps that could be tied to the next increase in minimum JRE version for VStar. We're currently being fairly modest in our requirements.

There may be other libs whose version could be increased. Apache Commons Math, for example. Last time I looked though, going up a major version entailed significant changes. Another issue entirely.

dbenn:
Hi @mpyat2, where are we at with this one? Was there a pull request?

mpyat2:
Hi @dbenn , 
There was no pull request.
If you don't mind replacing the library with a newer version, I'll create the pull request.

mpyat2:
On the other hand, there were some changes involving charting. Probably I'll try to create a new branch based on the current master.

dbenn:
Hi Max

Given that it’s a very specific replacement with no overlaps, plese feel free to make the change on master and don’t worry about the PR.

David

dbenn:
Oh, in that case, please feel free to do it all on a branch with a PR. :)

PR #133, opened 2021-03-18
Issue 117 jfreechart update

mpyat2:
Issue #117: JFreeChart library updated to 1.5.2


dbenn (commented):
Hi Max. Looks good but a question and a potential issue.

**Question**: is `public Dimension getPreferredSize()` a requirement of jfreechart 1.5.2 vs the old version? Are the width and height values chosen for particular reasons or just because something is needed and they seem reasonable in general?

**Issue**: I see a WWZ error dialog for the same dataset (a Kepler FITS file; will attach) with min period 0, max period 1, resolution 0.0001, everything else default in WWZ dialog) that I don't see with old jfreechart version. The data lends itself to a known problem in the WWZ implementation re: large gap in data. You end up with periods of 0 and so infinite frequencies (see top hits pane). The error occurs when you click the Frequency vs Time tab, presumably because of some difference in jfreechart's response to infinite values. Not a showstopper, just something to consider here and perhaps a new ticket.

Screen Shot 2021-03-20 at 08 58 15

[kplr010789273-2013131215648_llc.fits.zip](https://github.com/AAVSO/VStar/files/6174362/kplr010789273-2013131215648_llc.fits.zip)

mpyat2:
Hi David,
'public Dimension getPreferredSize()' is for compatibility with default sizes of the old JfreeChart (1.0.xx).
Without it the default chart size is quite big:
image

getPreferredSize() sets the same size as for JfreeChart 1.0.xx:
image

Issue IllegalArgumentException: indeed, JFreeChart ignored the infinite axis value in ver 1.0.13. Check for the infinite axis size already exists in ver 1.0.19 (the last 1.0.xx version):

@Override
    public TickUnit getCeilingTickUnit(double size) {
        if (Double.isInfinite(size)) {
            **throw new IllegalArgumentException("Must be finite.");**
        }
        this.power = (int) Math.ceil(Math.log10(size));
        if (this.integers) {
            power = Math.max(this.power, 0);
        }
        this.factor = 1;
        boolean done = false;
        // step down in size until the current size is too small or there are 
        // no more units
        while (!done) {
            done = !previous();
            if (getTickSize() < size) {
                next();
                done = true;
            }
        }
        return new NumberTickUnit(getTickSize(), getTickLabelFormat(), 
                getMinorTickCount());
    }

This code is the same as in ver 1.5.2

mpyat2:
@dbenn , please see the last commit to the issue-117-jfreechart-update branch.
This is a "cosmetic" fix. At least the exception does not occur immediately while switching tabs.
The infinite values still present so if the user chooses "Auto Range->Range Axis" from the pop-up menu, the exception occurs (and there are other ways to raise the exception).
It would be good to make changes in WWZ calculations to avoid infinity, for example, start calculations from 0+step if Minimum Period = 0 or do not allow the user to enter 0 as Minimum Period (with some warning message). This is for another branch of course.

dbenn (approved):
All looks good Max. I'm happy to merge this.

dbenn:
I'll create another issue for this. There are multiple tix on WWZ I think; perhaps some from SourceForge to transfer yet.

mpyat2:
Thank you!

Issue #138, opened 2021-03-26
VeLa: trouble with a signed positive exponent


dbenn:
From Max in email:

VeLa has troubles with a signed positive exponent.

You can check this with `Tools`->`VeLa`. 

Try entering `3.5754842e+18` and run the interpreter. You will see `20.718281828459` as a result! 

Without the `+` sign, all is OK: `3575484200000000000`.

dbenn:
One approach here is to intercept this before VeLa sees it via regex and change the string. I'd rather fix it via ANTLR4 though.

dbenn:
I'm actually thinking that there's no real need for E+nn in a real number when Enn can be used.

I've tried reworking the VeLa grammar to handle E+ but there's some issue I'm not seeing. There's only one unit test in `VeLaTest` affected by this and there's no real need for E+, as indicated.

dbenn:
I'm actually thinking that there's no real need for E+nn in a real number when Enn can be used.

I've tried reworking the VeLa grammar to handle E+ but there's some issue I'm not seeing. There's only one unit test in `VeLaTest` affected by this and there's no real need for E+, as indicated.

mpyat2:
@dbenn , I agree. There is no need to implement it. Currently, format procedures generates numbers without +

Issue #92, opened 2020-11-22
Make model equation term precisions consistent (use preferences)


dbenn:
From https://sourceforge.net/p/vstar/bugs-and-features/591/

Fourier and polynomial models are not consistent.

This does still seem to be the case. Perhaps one uses Preferences and the other does not. What about Loess? Any others?

mpyat2:
It's hard to make polynomial and Fourier coefficients fully precision consistent. The range of Fourier-model coefficients is relatively small; polynomial coefficients may vary by dozens of magnitude orders.
We could use "scientific" (%E) or "general" (%G) formatting for the polynomial coefficients and remain existing (decimal) formatting for the Fourier-model coefficients.

Issue #14, opened 2020-05-30
Model dialog function tab for R zeroPoint shows comma in value for non-English locales 


dbenn:
This is fine for VeLa, which is locale-sensitive, but not other languages.

However, while the VeLa function shows a locale-sensitive zeroPoint, the function's coefficients do not show as  locale-sensitive.

dbenn:
The situation is different for polynomial fit and Fourier models. For the latter, coefficients contain commas for all functions (R, Excel, VeLa) but not for the former.

PR #139, opened 2021-03-26
Issue 14 rcomma non en locales

mpyat2:
Hi David @dbenn,

This branch addresses issues #14 and #92.
The following files/classes have been changed:
ApacheCommonsLoessFitter.java
ApacheCommonsPolynomialFitCreatorPlugin.java
PeriodAnalysisDerivedMultiPeriodicModel.java
PeriodFitParameters.java

toRString(): locale-independent formatting (using ENGLISH locale)
toExcelString(): formula separator should be locale-dependent. If ',' is a decimal separator, the Excel formula separator is changed to ';'.

ApacheCommonsLoessFitter, ApacheCommonsPolynomialFitCreatorPlugin: polynomial coefficients are formatted now using scientific notation (with configured precision: in this case it is a total number of significant digits) via new functions NumericPrecisionPrefs.formatPolyCoef/formatPolyCoefLocaleIndependent. This is because the coefficients could have a huge range and must be formatted with exponential notation (at least in some cases).

Before the fix, polynomial coefficients were converted to strings implicitly (with implicit Double.toString() call), so they did not respect locale at all.

Excel formula issue (locale-dependent list separator) arose because the Excel formulas are implemented with SUM() function. We cannot use ',' when the decimal separator is also a comma (in my locale with ',' as a decimal separator, the default list separator is ';'). Why not get rid of SUM(a, b, c ...)? The formula could be in the form a + b + c +..., like other ones. What do you think?

Apparently, there is also a problem with functions for models generated by Loess plug-in. First, they are too long. And it seems they are incorrect. Maybe I'm doing something wrong, yet I cannot regenerate the model with those functions...

Concerning #138: formatPolyCoef returns string representation of a number in scientific notation with E+replaced by E as a workaround.

### DIFF
--- a/plugin/src/org/aavso/tools/vstar/external/plugin/ApacheCommonsLoessFitter.java
+++ b/plugin/src/org/aavso/tools/vstar/external/plugin/ApacheCommonsLoessFitter.java
@@ -37,6 +37,7 @@ import org.aavso.tools.vstar.ui.model.plot.StandardPhaseCoordSource;
 import org.aavso.tools.vstar.util.locale.LocaleProps;
 import org.aavso.tools.vstar.util.model.IModel;
 import org.aavso.tools.vstar.util.model.PeriodFitParameters;
+import org.aavso.tools.vstar.util.prefs.NumericPrecisionPrefs;
 import org.apache.commons.math.MathException;
 import org.apache.commons.math.analysis.interpolation.LoessInterpolator;
 import org.apache.commons.math.analysis.polynomials.PolynomialFunction;
@@ -139,15 +140,16 @@ public class ApacheCommonsLoessFitter extends ModelCreatorPluginBase {
 
 						double constCoeff = 0;
 
+						// We use formatGeneral because of a huge range
 						for (PolynomialFunction f : function.getPolynomials()) {
 							double[] coeffs = f.getCoefficients();
 							for (int i = coeffs.length - 1; i >= 1; i--) {
-								strRepr += "    " + coeffs[i];
+								strRepr += "    " + NumericPrecisionPrefs.formatGeneral(coeffs[i]);
 								strRepr += "t^" + i + "+\n";
 							}
 							constCoeff += coeffs[0];
 						}
-						strRepr += "    " + constCoeff;
+						strRepr += "    " + NumericPrecisionPrefs.formatGeneral(constCoeff);
 						strRepr += "\n}";
 					}
 
@@ -163,16 +165,17 @@ public class ApacheCommonsLoessFitter extends ModelCreatorPluginBase {
 
 						double constCoeff = 0;
 
+						// We use formatGeneral because of a huge range
 						for (PolynomialFunction f : function.getPolynomials()) {
 							double[] coeffs = f.getCoefficients();
 							for (int i = coeffs.length - 1; i >= 1; i--) {
-								strRepr += coeffs[i];
-								strRepr += "*A1^" + i + ",\n";
+								strRepr += NumericPrecisionPrefs.formatGeneral(coeffs[i]);
+								strRepr += "*A1^" + i + NumericPrecisionPrefs.getExcelFormulaSeparator() + "\n";
 							}
 							constCoeff += coeffs[0];
 						}
 
-						strRepr += constCoeff + ")";
+						strRepr += NumericPrecisionPrefs.formatGeneral(constCoeff) + ")";
 					}
 
 					return strRepr;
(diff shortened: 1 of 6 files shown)


src/org/aavso/tools/vstar/util/prefs/NumericPrecisionPrefs.java
```diff
@@ -156,10 +187,95 @@ public static DecimalFormat getOtherOutputFormat() {
...
+			s += ".0";
+		}
+		return s;
+	}
+*/	
+	
+	// Excel formula separator (list separator)
+	// There is no way to get system "list separator" (Windows specific
+	// locale-sensitive parameter). So we make an assumption. 
+	// Why not replace SUM() in Excel formulas with a simple a+b+c... expression?
```
dbenn, line 267: this seems like a reasonable approach @mpyat2

src/org/aavso/tools/vstar/util/prefs/NumericPrecisionPrefs.java
```diff
@@ -156,10 +187,95 @@ public static DecimalFormat getOtherOutputFormat() {
...
+	public static String formatPolyCoefLocaleIndependent(double num) {
+		//return formatGeneralLocaleIndependent(num);
+		return formatScientificLocaleIndependent(num);
+	}
+	
+	// Scientific format
+
+	private static String formatScientific(double num) {
+		// .replace("E+", "E"): VeLa issue workaround
+		return String.format(Locale.getDefault(), "%." + otherDecimalPlaces + "E", num).replace("E+", "E");
```
dbenn, line 229: a reasonable workaround until I fix #138 @mpyat2

src/org/aavso/tools/vstar/util/prefs/NumericPrecisionPrefs.java
```diff
@@ -156,10 +187,95 @@ public static DecimalFormat getOtherOutputFormat() {
...
 	}
 
+	// PMAK (2021-03-25): 
+	// Special case: format for polynomial coefficients.
+	// These coefficients have a huge value range and cannot be satisfactorily
+	// represented by DecimalFormat.
+	// What format would be better: general (%.xxG) or pure scientific (%.xxE)?
+	// Both have issues (for the current VeLa numeric parser):
+	//   1) signed positive exponent cannot be correctly parsed by VeLa now (i.e. 1.234E+15)
+	//   2) long number strings without decimal separator cannot be parsed too (i.e. -608516245008941)
```
dbenn, line 209: @mpyat2 Thanks. I'm adding NumberParser unit tests to look at this. I'll either address it in #138 or in a separate issue. VeLa integers are currently Java ints. Should be longs. For now, the workaround would be `-608516245008941.0` to force VeLa to parse it as a real.

plugin/src/org/aavso/tools/vstar/external/plugin/ApacheCommonsLoessFitter.java
```diff
@@ -139,15 +140,16 @@ public String toString() {
...
 						double constCoeff = 0;
 
+						// We use formatGeneral because of a huge range
 						for (PolynomialFunction f : function.getPolynomials()) {
 							double[] coeffs = f.getCoefficients();
 							for (int i = coeffs.length - 1; i >= 1; i--) {
-								strRepr += "    " + coeffs[i];
-								strRepr += "t^" + i + "+\n";
+								strRepr += "    " + NumericPrecisionPrefs.formatPolyCoef(coeffs[i]);
+								strRepr += "*t^" + i + "+\n";
```
dbenn, line 148: Thanks @mpyat2. This was a hang-over from when functions were not VeLa, merely "descriptive" and I obviously did not test this transition properly for Loess models.

dbenn (commented):
Thanks for your work on this so far. See my specific comments re: code lines.

src/org/aavso/tools/vstar/util/prefs/NumericPrecisionPrefs.java
mpyat2: ok

src/org/aavso/tools/vstar/util/prefs/NumericPrecisionPrefs.java
mpyat2: ok

mpyat2:
I've created issue #145 for the ApacheCommonsLoessFitter model problem

mpyat2:
In the last commit, I've removed Excel =SUM() function and got rid of Excel list separator (not used anymore).
I've also replaced incorrect ApacheCommonLoessFitter model functions by Mediator.NOT_IMPLEMENTED_YET text.
A new issue  #145 for the ApacheCommonsLoessFitter has been created.

dbenn:
@mpyat2 This one looks close. Anything remaining from your POV? If not, we could merge.

mpyat2:
@dbenn Yes, please merge

Issue #141, opened 2021-03-28
VeLa integers should be long not int


dbenn:
Since VeLa has a single integer type, it should allow for the largest range.

Right now it's `int` so should be `long`.

In #139 for example, a large integer model coefficient is failing to parse as expected because it is outside of the allowed integer range.

PR #144, opened 2021-03-28
#141: changed VeLa integers to be Java longs

dbenn:


mpyat2:
`	/**`
`	 * Get the integer value from the text field, if possible, otherwise return null`
`	 * if no valid number is present in the textField.`
`	 * `
`	 * @return The double value or null.`
`	 */`
`	public Integer getValue() {`
`		Integer value = null;`


Probably should be `@return The integer value or null.`

The same in LongField.java: should be @return the long instead @return the double

Everything else looks good.

mpyat2 (commented):
`/**`
`* Get the integer value from the text field, if possible, otherwise return null`
`* if no valid number is present in the textField.`
`*`
`* @return The double value or null.`
`*/`
`public Integer getValue() {`
`Integer value = null;`

Probably should be `@return The integer value or null.`

The same in LongField.java: should be @return the long instead @return the double

Everything else looks good.

dbenn:
Thanks @mpyat2. I corrected the comment in Long and Integer field classes as well we checking for null in these classes and `Checkbox` `setValue()` methods (as you already did in `DoubleField`).

Issue #140, opened 2021-03-26
Simple text format: observation flags are not consistent between reading and writing.


mpyat2:
SimpleFormatObservationSinkPlugin writes "G" observation flag for "good" observations. In turn, SimpleTextFormatValidator() acceps only "D" flag ("discrepant").
So, simple text format, generated by VStar ["Observations": File->Save (Type=Simple)] cannot be read as "Download or Simple" source because all "good" lines contain unrecognizable "G" flag!

dbenn:
Wow. This is odd @mpyat2. Given that we are writing observations that may have been loaded as AAVSO Download format in simple format, either the latter should expand to allow validation flags other than D (in the early VStar spec) or not output a validation flag at all.

Thoughts?

@SaraAAVSO, do you have a view on what is best to do here, given your recollections of the early VStar development period vs how things have moved on. I recall an evolution of validation flags in AAVSO also.

dbenn:
@SaraAAVSO: some additional background. Until quite recently (the last release in 2020), saving obs was a special feature, whereas now there's a plugin category called observation sink (vs source). Internally, in the last release I converted the existing save obs code to use the new plugin interface, and the issue Max raises crept in during that conversion.

SaraAAVSO:
Hi @dbenn and @mpyat2,

I don't remember how we wound up with validation flags of "B" and "D" which as far as I know were never official validation flags. As explained in the Data Download tool: 

Validation Flag: This flag describes the level of validation of the observation.

    V means the observation has passed our validation tests.
    T means that during the validation phase it was flagged discrepant and should be used with extreme caution.
    Z means it has only undergone pre-validation, meaning it was checked for typos and data input errors only.
    U means it has not been validated at all and should be used with caution.

So "T" = bad while "V" and "Z" should be considered good. I'm not sure what we decided to do with "U" but I think it will be displayed along with other good data in VStar. 

If someone requests data using the Data Download Tool and asks not to get discrepant data, they will get only data with valflags "V" and "Z". If they request that discrepant data is included they will also get observations with the "T" and "U" flags.

I agree that the same format should apply to reading and writing.

Does this answer your questions?

Thanks, Sara

mpyat2:
So both "G" (good) and "D" (discrepant) are purely VStar specific values. Both these flags recognized if the input format is "Download", "G" rejected if the input format is "simple" however "D" is recognizable for the "simple" format.

[...]

PR #146, opened 2021-04-03
#140: using common valflags for simple and AAVSO download formats and added unvalidated valflag

dbenn:
This gives backward compatibilty with simple format (that permitted the D valflag) and unifies both formats with respect to valflags.


mpyat2 (approved):
Looks good. I'll change the FlexTextFormat the same way.

dbenn:
Thanks @mpyat2

dbenn:
@mpyat2 Unit tests pass. Just opening various files for further testing. Let me know if you see any problems otherwise I'll merge in the next day or so. Thanks.

SaraAAVSO (approved):

PR #150, opened 2021-04-27
Issue 143 models not recorded

dbenn:


mpyat2:
Hi @dbenn , now it works as expected!

Issue #155, opened 2021-05-10
Install/Update All buttons in plugin manager


dbenn:

dbenn:
@mpyat2 @SaraAAVSO I've taken the approach of removing the Delete All button, adding an All? checkbox which applies to Install, Update and Delete buttons.

<img width="510" alt="image" src="https://user-images.githubusercontent.com/4115628/118273930-adc79680-b503-11eb-9908-89272936b9cc.png">

mpyat2:
Thank you, @dbenn 
I'm looking at it and will add a couple of minor changes.

dbenn:
If you are happy with my last minor commit @mpyat2, and if we think we are done with UI changes, I'll merge the branch and update the user manual.

The Install and Update buttons have the same effect when All? is selected. We could hide the Update button in all-mode, but I'm not sure whether to be concerned about that. Thoughts?

mpyat2:
I think all is ok, see also my comment to #155.

PR #156, opened 2021-05-14
#155: removed Delete All button and added an All? checkbox to apply operations to Install, Update, and Delete buttons

dbenn:


mpyat2:
Hi @dbenn ,
I've made some changes. Please see the commit.
First, I've made Install/Update/Delete methods more self-consistent; made some cosmetic fixes, and added a custom renderer to the plugin list: now installed plugins are bolded.

dbenn:
Thanks for all the changes/fixes @mpyat2. I don't know why I had a loop in `setAllButtonStates()`! Thanks for spotting and fixing that. Upon reflection, it's clear that the method name was not right either, so I've changed that and the comment above the method.

mpyat2:
Hi @dbenn , since Update and Install do the same if 'All?' is in an active state, I've made a minor change that disables 'Update' when 'All?' is checked.

mpyat2:
Well, another minor fix (accessing ListModel in a background thread). I hope it's all for now.

dbenn:
Will allow a bit more time for us to test this out before merging.

mpyat2:
ok

mpyat2 (approved):

Issue #177, opened 2021-06-19
Test for JVM bitness while installing VStar under Windows


mpyat2:
If Java is 32bit while Windows is 64bit, the installer must detect it and set memory options accordingly

mpyat2:
New Windows installer checks for JVM bitness using 'JavaOsArch' helper.

PR #209, opened 2021-12-17
Issue 177 installer detect jvm bitness

mpyat2:
Hi @dbenn , the changes in this branch are related to the Windows installer only, see issue #177

Now the installer checks JVM bitness instead of assuming that JVM bitness = system bitness.
So I hope we will get rid of errors when the memory size is set to half of the system memory size while 32-bit java cannot consume it and refuses to start.

The last Sebastian's case proved that it works, however more extensive check is desirable. I tested it on one 32-bit machine (Windows 7), two machines with Windows 10 64-bit, and two machines with Windows 7 64-bit.


dbenn (approved):
Looks good. Thanks!

Issue #208, opened 2021-12-16
Extend "JD to Calendar Date" plugin


mpyat2:
Currently, the plug-in looks like a stub:
![image](https://user-images.githubusercontent.com/44919945/146399923-b131d4ff-8ca1-416e-85c4-13e749e909f8.png)

It would be better to have something like this:
![image](https://user-images.githubusercontent.com/44919945/146400393-21ee32fe-82bc-45b6-8cc1-2f225cc42040.png)

mpyat2:
It partially covers Issue #83. Issue #83 requests JD<->HJD conversion too, it requires coordinates.

mpyat2:
Thank you, @dbenn

PR #211, opened 2021-12-20
Issue 208 extend jd to cal date plugin

mpyat2:
Hi @dbenn ,

I've made a completely new "JD to Calendar date" plug-in (see Issue #208).
It is bidirectional, one can convert from JD to Date/Time and vice versa.
image

I also patched DoubleField.java and IntegerField.java. There was a bug in both classes that could cause Null Pointer Exception: I replaced null assignment to the variable by the 'return null' statement to avoid comparison of null with a variable.

Also, I've added 'catch (VeLaParseError e)' and 'catch (VeLaEvalError e)' in IntegerField.getValue() like in DoubleField.getValue(). Those exceptions can be thrown by NumberParser.parseInteger().

### DIFF
--- a/plugin/src/org/aavso/tools/vstar/external/plugin/JDToDateTool.java
+++ b/plugin/src/org/aavso/tools/vstar/external/plugin/JDToDateTool.java
@@ -243,13 +243,17 @@ public class JDToDateTool extends GeneralToolPluginBase {
 		
 		// Set form fields to values corresponding to the current time
 		private void setCurrentTime() {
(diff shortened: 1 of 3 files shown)


dbenn (approved):
Thanks for the NPE bug fix and the plugin overhaul Max!

dbenn:
I meant to add: the plugin is awesome. Tried it out. Thanks Max.

mpyat2:
Thank you, David!

Issue #222, opened 2021-12-29
VeLa function parameter binding should always occur in top-most scope


dbenn:
@mpyat2 asked a question about VeLa function parameters in email in the context of #221 re: why a model function `t` parameter always seems to be ignored, yet the model works as expected.

This took me on a bug hunt which led to the astounding conclusion that in the presence of a `VeLaValidObservation` object in the function scope stack, a parameter with an existing name will end up being bound to that object. `t` resolves to `time` in this case, since `t` is a synonym for `time` in `VeLaValidObservation` objects.

The parameter is _not_ bound to the function's activation record. It _is_ bound to the `VeLaValidObservation` object's variable cache as mentioned above, but lookup ultimately resolves to the `time` variable in the `ValidObservation` object that the `VeLaValidObservation` object encapsulates.

Function parameters should _always_ be bound to the current activation record, with no search down the stack of scopes (`VeLaScope`s (activation records) and `VeLaValidObservation`s), unlike we want for variable search/bind. I thought I _was_ always doing this, but apparently not.

There may be implications for `VeLaModelCreator` code, but that can the subject of another issue.

dbenn:
So, the problem has been fixed and all VeLa unit tests pass.

However, as anticipated, `VeLaModelCreator` _is_ affected.

An example will help to best explain this. Fourier models from DCDFT are created in the time domain, i.e. with the times of observations being the values assumed to be passed to the model function. This is not always true of other models. Of course, what this means is that if the model is applied in phase plot mode, the model creator plugin will assume phases, not times, as the values to be passed to the model function, which will yield erroneous model observations and residuals.

This is actually as it should be. We have simply been lulled into a false sense of security (actually an incorrect version of reality) because, as you found (and thank you!), the time (`t`) parameter to the model function was being ignored, in favour of the `t` binding in the current observations `VeLaValidObservation` object.

I hope this is making sense to you @mpyat2.

PR #226, opened 2021-12-29
#222: changed binding of function parameters to always be in function's activation record

dbenn:


mpyat2 (approved):
Now it works as expected: VeLa function call is sensitive to its argument (t). Changing locale to European with ',' instead of '.' leads to incorrect model evaluation, _as expected_.

PR #230, opened 2022-01-03
Issue 229 expressionvisitor parsedouble bug

mpyat2:
@dbenn ,

I made additional changes in ExpressionVisitor.parseDouble to make it more error-proof. Please see the comments in the source code. Could you please make unit tests for it?

I also checked if the locale affects _output_ formats used for generating textual representations of models.
The exponential format used for polynomial models (NumericPrecisionPrefs.formatScientific()) utilizes String.format() method.
Interestingly that it always uses normal Latin E for exponent regardless of locale (even for Ukrainian uk_UA while DecimalFormat.format prints Cyrillic Е for the scientific formats like "#.########E0" (Java ver>8), fortunately, we never use it in VStar).
So model expressions in the "Model information" are correct independently on the Java version and locale (I hope so).


dbenn (approved):
I think this looks good Max. Thanks!

Issue #227, opened 2021-12-30
Applying polynomial model using VeLa model creator plugin gives incorrect result


dbenn:
Try creating any polynomial fit then applying it via the VeLa model creator plugin.

mpyat2:
1) Load KP Lyn light curve (observer = PMAK) from AAVSO db
2) Polynomial fit degree= 15 for Johnson V
![image](https://user-images.githubusercontent.com/44919945/147731534-881b62c7-4b6d-49ec-bda3-5bec7381d456.png)
3) VeLa model function
zeroPoint is 2459332.35709

f(t:real) : real {
    3.513493389760E19*(t-zeroPoint)^15 +
    1.486885038629E18*(t-zeroPoint)^14 +
    -6.994358939253E17*(t-zeroPoint)^13 +
    -2.778248197535E16*(t-zeroPoint)^12 +
    5.705533522828E15*(t-zeroPoint)^11 +
    2.089863042692E14*(t-zeroPoint)^10 +
    -2.460853674112E13*(t-zeroPoint)^9 +
    -8.103385927143E11*(t-zeroPoint)^8 +
    6.037656131770E10*(t-zeroPoint)^7 +
    1.722865043295E09*(t-zeroPoint)^6 +
    -8.456524341291E07*(t-zeroPoint)^5 +
    -1.993607493521E06*(t-zeroPoint)^4 +
    6.522979829427E04*(t-zeroPoint)^3 +
    1.170687252721E03*(t-zeroPoint)^2 +
    -2.269285161529E01*(t-zeroPoint)^1 +
    1.033057728586E01
}

3) Reload KP Lyn observer PMAK
4) VeLa model for Johnson V
![image](https://user-images.githubusercontent.com/44919945/147731718-597065a2-70f1-4a70-988e-0e6af3619811.png)
4) Model curve
![image](https://user-images.githubusercontent.com/44919945/147731786-0a0f0a25-2280-464b-b15f-59c951946eda.png)

The curves look the same, values (model pane) differ in the 5th digit after the decimal point (it's ok)

I used fresh master brunch

dbenn:
Okay, thanks Max. I'll check that I can reproduce this and if so, see if I can still reproduce the original case I saw this in.

dbenn:
Very strange. I checked out master and then did a git pull. Here's what I see:

[...]

Issue #120, opened 2021-02-16
Unit Tests for Plugins


yorkpf:
Plugins deserve to have their own unit tests, as currently exist for core code. Especially for Observation Source Plugins, rather than test these manually. Some plugins have existing UTs, but only a minority.

dbenn:
One approach here is to add the infrastructure in the class that sits at the base of all plugins via JUnit's test case, requiring only a test method to be implemented. A default no-op implementation of the test method could be created for backward compatibility. Forcing it to be implemented would mean the developer makes a deliberate choice not implement anything but an empty test though.

dbenn:
The AoV plugin would be a good first candidate to try this on, especially since a design-for-test would require test code to bypass UI code while testing the key parts of the plugin code. It would also more easily give us confidence that a multi-core version still worked as expected.

dbenn:
The base class of all plugins should have a `Test` method that returns either `null`, `true` or `false` or an `Optional<Boolean>` to indicate that there is no test, the test passed, or the test failed. From a purely intentional and slightly aesthetic viewpoint, I favour `Optional<Boolean>` since it connotes a freedom to choose whether or not to write a test.

dbenn:
Further, we will add a unit test whose job it is to load each plugin, invoke its test method, and report the result of each non-null/optional test.

dbenn:
Okay, plugin unit test workflow is working now. Had to recall what the issue was from #6 and also pay a bit more attention to the exception in the GitHub action log.

dbenn:
Further to the last comment, as per commits, there was a bit more to be done.

What we now have is a generalised, separate test runner for currently installed plugins. This is useful during development as well as in GitHub actions.

The way it works is that a plugin developer can opt to write a test by overriding the `test()` method in a plugin and all such tests will be for internal plugins and and any that can be installed via `plugin/build.xml`'s test target (`ant test` in `plugin` directory). An overall pass/fail and individual report of tests is given on `stdout`. If not overridden, the `test()` method returns `null` signalling to the `PluginTest` test runner class that there is no test for that plugin. Otherwise, true or false is returned to denote pass/fail.

@mpyat2 I'd be interested in your thoughts on all of this. My aim here is to make writing plugin tests more integral to the plugin development process while adding minimal overhead to each plugin.

dbenn:
See here for an example:

[...]

PR #232, opened 2022-01-13
Issues 120 UTs for plugins

dbenn:
There are two main components here. One is a general JUnit based mechanism for plugin unit tests, a sort of design-for-test approach. The second is improvements to the actions (workflow) for running unit tests for plugins (added) and multiple versions of Java. The latter was triggered by #227.

See https://github.com/AAVSO/VStar/actions/runs/1692722608 

More plugin unit tests can be added in future but these can be handled as separate issues. Only 3 out of 40 plugins currently have unit tests.

There used to be a few standalone plugin UTs, but now there's one JUnit test class that invokes a test method in all plugins, as per #120.

There are other related changes to Ant, YAML (GitHub Actions CI), and a couple of tests (HJD) that failed with different versions of Java (>=11) relating to floating point representation >= the 10th decimal place. There are also a few random format changes not directly related to the issue at hand.

@yorkpf: comments welcome, but don't feel you need to get down into the weeds unless you want to.
@mpyat2: please review the code changes and Java versions in the YAML.


dbenn:
@mpyat2 one thing to be considered is what list of Java versions to test each time. In the two YAML files under `.github/workflows`, we currently have:
~~~
java: [ '8', '9', '11', '14', '15', '16' ]
~~~

mpyat2 (approved):

dbenn:
> Well, the difference is negligible. I see the same behavior under Windows (Java 8 vs Java 16). I have not run all the tests yet made a small test class and copied testMeanObliquityLowPrecision1 code.
> 
> Here are some links that could shed light on the cause: https://stackoverflow.com/questions/25404088/math-pow-yields-different-result-depending-on-java-version https://stackoverflow.com/questions/4232231/whats-the-difference-between-java-lang-math-and-java-lang-strictmath https://en.wikipedia.org/wiki/Strictfp

Agreed. That's what I was thinking too (negligible). Thanks for the links. I came across `strictfp` differences as well, but not `pow` yet. That could be a factor with one of the tests (`testObliquity1`) which indirectly invokes `meanObliquityHighPrecision` that makes use of `pow`. Not sure about the other one.

In any case, looks like we're OK to close this. Thanks for the review.

Issue #194, opened 2021-10-31
Checkpointing in VStar


yorkpf:
We need some way of checkpointing our activities in VStar, so that we can save our work in case of a mistake, or between different sessions  of VStar, and pick it up again from that same point later on. 

A typical scenario is this: You load up some ZTF data, work on the zr and zg light curves for a while, eliminating bad catflag data, discrepant obs, outliers,etc. This all takes care, time and effort. You then load up some ASAS-SN data, but forget to tick the “Add to current?” box. Bang! All of the current ZTF work vanishes and you have to repeat it. Painful. I have had as many as 4 or 5 edited light curves on screen at once and lost the lot. 

As I think further on this, it would be nice simply to save our work from the current session so that further work can be done on it in a future session, without going back to square one.

A possible solution (and maybe easier) to the problem of over-writing existing work by accident might be to pop up a dialog box, asking "Do you really want to load a new obs dataset over the current one? [David Benn]

dbenn:
The idea of a "project" has been considered in the VStar context for a long time. This may be appropriate for the current problem. Need to spend time ruminating.

yorkpf:
The “project” idea sounds OK, if I understand it correctly

PR #233, opened 2022-01-14
Issue 221 vela observation source plug in

mpyat2:
Known issue: The plug-in still cannot be used effectively in WebStart mode due to a lack of "file open" and "paste from clipboard" capabilities (security restrictions).


dbenn:
Hi @mpyat2. Just looking at the pull request. Can you tell me more about VeLa XML files and the context here? Thanks.

mpyat2:
In addition to the VeLa code, the plug-in requires parameters Minimum JD, Maximum JD, number of points to calculate, and type of Julian date (JD/HJD/BJD).
These additional parameters are saved along with the VeLa code in an XML file, see an example attached.

While loading a file, the plug-in tries to interpret it as XML, if failed, it loads it as plain text, so it is possible to load simple text files with VeLa code (like the VeLa model plug-in does).

@dbenn , please do not merge this branch with master now, I should test some possible encoding issues.

dbenn:
Thank you for all of this @mpyat2! I'm starting to look at this.

dbenn (approved):
This all looks good Max. I'm wondering whether `.vlx` could/should be something like `.vmx` only because it's specifically for a VeLa model with additional data other than just VeLa code. I'm not too worried either way though.

As an aside, this kind of XML may help to inform the kind of "project" idea floated in #194.

dbenn:
@mpyat2 I was going to create an issue to use one of your new save functions but notice at least in my English locale that it is saving and loading OK; I will at least fix the lambda symbol though as per #234

mpyat2:
Thank you, @dbenn !
Concerning extension: I'm not sure what is better. Both extensions are used already by other products (it's hard to find something unique now). vmx = VMware Virtual Machine Configuration, "VLX files mostly belong to AutoCAD by Autodesk".
Probably, if we invent 4-char extension for it it could be unique... Currently, I have no idea. Anyway, we can change it anytime (before the release).

dbenn:
Re: your 4 letter extension idea: `.vlmx`?

dbenn:
Re: Ukranian comment, so sounds like we need an issue to fix this then.

mpyat2:
@dbenn , when you have time, could you please merge this branch?

PR #237, opened 2022-01-23
Issue 236 UTs

dbenn:


mpyat2 (approved):
Looks ok. I see that there are 'mysterious' input locales that cannot be managed reasonably. Well...

dbenn:
With the nn fix committed and your review, I'll merge this PR.

PR #238, opened 2022-01-24
Issue 235 vela dialog save

mpyat2:
1) Unified load/save procedures  in VeLaDialog, VeLaObservationFilterDialog and VeLaObSource. 
2) Removed unnecessary VeLaFileLoadChooser, VeLaFileSaveChooser (replaced by LoadChooser/SaveChooser with extended functionality).
This point was removed. See my comment below. ~~3) A new overloaded TextDialog constructor~~
```
public TextDialog(String title, List<ITextComponent<String>> fields, boolean show, boolean scrolled)
```
~~if scrolled == true, the text fields get enclosed in scroll boxes.~~

~~See how the VeLa dialog behavior changed:~~
~~a) old version, after opening long code (user needs to resize the dialog to find the buttons):~~
image
~~b) new version with scroll boxes:~~
image


mpyat2:
Well, it seems I need to make a regression: Point 3 (scrollable text) works quite strange in some cases.

mpyat2:
I removed the changes related to point 3 (anyway those changes were not related to the fix).

However, I plan to return to it. Scrollable boxes look better. The problem is in VeLaDialog static members (codeTextArea, resultTextArea).

dbenn (approved):
Looks good, thanks Max. Thanks for your scroll bar fix attempts. We can defer this to another issue.

PR #240, opened 2022-01-25
NumericPrecisionPrefs output format compatible with ExpressionVisitor input one.

mpyat2:
Output DecimalFormatSymbols are now the same as input ones.

Besides DecimalFormat, NumericPrecisionPrefs uses String.format() for values with exponents (used in polynomial models). Interestingly that String.format() does not suffer from Java version-to-version incompatibility like DecimalFormat. Although this is an inconsistency within the Java libraries themselves.


dbenn (approved):

mpyat2:
Thank you, @dbenn . I'm not sure that it is the last locale-specific issue however I hope there will not be many... Until new refreshment of Java library :)

dbenn:
Indeed @mpyat2 :)

PR #241, opened 2022-01-26
#138: removed E+nn from real numbers in VeLa grammar; removed UT test…

dbenn:
… case for this; also removed + as a valid numeric prefix


mpyat2 (approved):

dbenn:
See also additional commit to remove + prefix from number parser text cases which caused CI to fail. Snuck in removal of + prefix before numbers because it's not necessary.

PR #247, opened 2022-04-11
246 B-V observation source plugin creates observations with no names

dbenn:
Well, this turned out to be bigger than Ben Hur. The name is being set now for B-V observations. In addition, the mag uncertainty is now being computed and set. Finally, a test method was added for the plugin, something I plan to do when the opportunity arises because the test coverage for plugins is low.

@mpyat2, if you are happy to review this, that would be great.

### DIFF
--- a/plugin/src/org/aavso/tools/vstar/external/plugin/BMinusVObservationSource.java
+++ b/plugin/src/org/aavso/tools/vstar/external/plugin/BMinusVObservationSource.java
@@ -21,7 +21,6 @@ import java.awt.Color;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


mpyat2:
Hi @dbenn ,
I propose adding the following line (marked yellow):
image

Initially, bvOb.getJDflavour() == JDflavour.UNKNOWN. For UNKNOWN values, AbstractObservationRetriever.categoriseValidObservation() sets observation's jDflavour to the ObjervationRetriever default value that is JD by default.

We can assume that the time mode for both B and V bands is the same and take JDflavour from V observations.

mpyat2 (approved):

mpyat2:
Thanks David!

Issue #51, opened 2020-08-25
User defined series need to use copies of original observations


dbenn:
Although one can create a new series from a filtered series, the observations are not cloned, so they retain the same band etc as the original. This means that operations such as polynomial fit will report the title of the operation in models dialog as whatever the underlying observations band was. Other operations, such as observation transformation (e.g. mag baseline shifter plugin) will also operate on the original data.

dbenn:
Both filtered series and new series created from a filtered series should be cloned.

Also, for another changes to `ValidObservation` objects, copy-on-write semantics could be considered.

dbenn:
We don't necessarily need to be overriding `@clone`, especially since it returns `Object`. We can instead create a `copy()` method that returns `ValidObservation`.

dbenn:
My inclination here is to say that since a user defined series will be created for the purpose of analysis of some kind, quite possibly with some important modification first, e.g. a baseline shift may be applied, a polynomial fit may have been created and the residuals used as the basis of a new series, that we should not feel compelled to slavishly copy every piece of information from`ValidObservation`.

Instead, time, mag, error, heliocentric status are all that may be required. I was going to add band to this list, but we have created a new series from what could possibly be a number of observations that have different bands. The wisdom of doing so could be argued, but still, it's possible.

Accordingly, I propose to add a member to `ValidObservation` which is a reference to the `ValidObservation` from which the new observation arises. Under normal circumstances, this would be `null` or if `Optional<ValidObservation>` is the type, non-existent. Analysis code would then be free to access the originating observation from the one in the new series.

One advantage is that creation of the new observation will be faster than a full clone.

dbenn:
Will add a method such as `minimalObservation()` for this purpose.

dbenn:
In the end I have decided that a full copy was not too onerous so have implemented `copy()`.

PR #248, opened 2022-04-15
#51: implemented full ValidObservation copy

dbenn:
Hi @mpyat2, I have implemented a full `copy()` method, used it in `FilteredObservationMessage` and `SeriesTypeCreationDialog` and tested it with mag baseline shift so far. I've also added a `ValidObservation` unit test for `copy()` and mutability.

If you have a look at this, it would be appreciated.

### DIFF
--- a/src/org/aavso/tools/vstar/data/CommentCodes.java
+++ b/src/org/aavso/tools/vstar/data/CommentCodes.java
@@ -17,7 +17,6 @@
  */
 package org.aavso.tools.vstar.data;
 
-import org.aavso.tools.vstar.data.CommentType;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


mpyat2 (approved):

mpyat2:
Very useful change. Now filtered and new series created from filtered are really independent of each other. However, the magnitude baseline shifter still does not show new series created from filtered ones.

dbenn:
> Very useful change. Now filtered and new series created from filtered are really independent of each other. However, the magnitude baseline shifter still does not show new series created from filtered ones.

Thanks Max. There is an issue and branch in which a create series menu item was proposed and added. Now I just need to find the issue. :)

Issue #49, opened 2020-08-24
Magnitude baseline shifter observation transformation plugin does not show all series 


dbenn:
Magnitude baseline shifter observation transformation plugin does not show all series in series selection dialog.

For example, if more series are created from filters. Also doesn't show synthetic series, e.g. model, residuals, means.

dbenn:
This will also be true for all observation transformation plugins.

dbenn:
This needs to be fixed in conjunction with #5 otherwise there will be side effects on objects included in a filter or user defined series.

dbenn:
Notice that the difference between the need to copy observations in this context is different than say, in the AoV plugin. In the latter case, we only need to copy mag and time, whereas in the current context, the whole observation object should be copied in toto.

Issue #84, opened 2020-11-22
HJD conversions should be undoable edits


dbenn:
We have a generic undoable edit framework in VStar. Right now only observation exclusions and transformations are undoable edits. HJD conversion is a good candidate for this as well.

dbenn:
Indeed, rewriting the HJD converter as an observation transformer plugin would not only seem reasonable, it would give us undoable edits for "free"!

dbenn:
See also Max's comment in #249 (https://github.com/AAVSO/VStar/pull/249#issuecomment-1100834035) re: new series being ignored by the current HJD converter. If the HJD converter is rewritten as an observation transformer, as suggested above, the user will be asked for which series to convert, including user defined series, so this problem would be addressed in that way.

dbenn:
For the undo/redo actions, either JD/HJD deltas could be used with the appropriate `UndoableActionType` operation specified or the pre-conversion values could be stored. Either way, real numbers will need to be stored.

PR #249, opened 2022-04-15
Issue 195 label light curve

dbenn:
@mpyat2 further to #51 which we've just fixed (ob copy method), this branch adds a Create Series View menu item. Series created via this can then be subject to observation transformation operations, such as mag baseline shifter.

Then, once this has been reviewed and merged, I can move on to #49 to allow more series to be selected from for series creation, including the Filter series.


mpyat2:
Hi @david, I see a strange behavior: the new series is not visible in the 'Observations' pane. Please see the video attached: https://www.youtube.com/watch?v=TNOLVJMtnEA

dbenn:
> Hi @david, I see a strange behavior: the new series is not visible in the 'Observations' pane. Please see the video attached: https://www.youtube.com/watch?v=TNOLVJMtnEA

Thanks @mpyat2. Hmm. Yes. The answer will be in `VisibleSeriesRowFilter` I think. I did notice this awhile ago and then neglected to capture it! Thanks for picking it up!

dbenn:
Please review the latest changes from that last commit @mpyat2. This seems to work.

mpyat2 (approved):

mpyat2:
Thank you, @dbenn , it works now.

mpyat2:
Hi @dbenn ,
New series created via "Create Series" are independent of the loaded data now. However, they behave differently. For example, they are not visible for the "Heliocentric JD converter plugin" (it takes observations from getLatestNewStarMessage). I'm not sure it is a bug, rather a feature (?).
Also, those new series seem not visible for filters created via "Simple filter...", "Vela Filter...", "Observer list filter...", and, probably, other filters (however, "Filter from plot" takes them).
Again, I'm not sure it is a "bug". What do you think? At least this behavior should be documented somewhere.

dbenn:
Hi @mpyat2. Indeed. Thanks for this! I have captured this in #84 and a new issue, #253. I'm not sure whether the filter mechanism is a bug. I suspect it is, but it needs some thought/conversation, whereas I _do_ think the HJD behaviour is a bug since I can't see a good reason to disallow a user defined series from being kept separate re: HJD status from its original observations.

PR #251, opened 2022-04-16
Build time in AboutBox

mpyat2:
Hi @dbenn , I've added a timestamp along with the existing git revision number: the timestamp is the build time in "yyyy-MM-dd HH:mm" format. The time is local, it would be better to have UTC yet I have not found if it is possible using the standard ant tstamp task.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -47,7 +47,9 @@
 
 	<target name="init">
 		<!-- Create the time stamp -->
-		<tstamp />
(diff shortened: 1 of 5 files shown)


dbenn (approved):
Looks good Max. Thanks. YYYY-MM-DD is good and as I have commented whimisically elsewhere, I find other date formats confusing. :)

Issue #258, opened 2022-04-29
Time error estimate of minimum or maximum from a polynomial fit


dbenn:
See Roy Axelson's request here: https://www.aavso.org/comment/162089#comment-162089

dbenn:
Grant Foster's light curve analysis book briefly talks about uncertainty of minima/maxima when using a model, such as binned means or polynomials, but unlike his treatment of period error for which standard error of the frequency and FWHM are suggested, there is not a similar concrete suggestion for polynomials fits.

dbenn:
This interesting overview of methods by Tom Richards of Variable Stars South focusses on polynomial fitting and Kwee van Woerden and on balance seems to favour polynomials, especially low order polynomials:

https://www.variablestarssouth.org/wp-content/uploads/2020/01/nl_2020-1.pdf

It emphasises the importance of "...estimate which order of polynomial fits best without over-fitting, and what does that do for the vital polynomial curve-fit to minimum" (which VStar already provides, via RMS, AIC, BIC goodness of fit measures) and minima/maxima error estimates but doesn't examine that in any detail at least for polynomial fits. It also suggests the utility of a suite of methods, as found in Bob Nelson's `Minima 25`.

What's needed is equations, algorithms or source code though.

dbenn:
This paper (via Brad Walter) is useful re: polynomial uncertainty:

https://www.researchgate.net/publication/227023900_On_Methods_for_the_Light_Curves_Extrema_Determination

dbenn:
See also https://www.aavso.org/tmaxmin which also provides some commentary on this, including mention of Bob Nelson's Minima software.

I also want to make note of the passing of Bob Nelson on November 2 2022 (3 days ago).

Issue #256, opened 2022-04-25
DCDFT FWHM as a measure of period error


dbenn:
FWHM, the period range near the peak over which the periodogram power is at least half its peak value.

See #255 for background.

I'm keeping these issues separate from an implementation viewpoint because they are very different with respect to how they will be handled within VStar.

dbenn:
Make FWHM available for a selected peak period via a `FWHM` button and also in the Models dialog for comparison with the standard error values.

dbenn:
Show as lower and upper bound of FWHM as per Grant's R code, e.g.
~~~
         freq    power      amp      per      freq.se     amp.se    fwhm.lo    fwhm.up
46 0.00299914 156.4582 2.825638 333.4292 2.532946e-06 0.05835603 0.00289931 0.00308787
     per.lo   per.up
46 323.8478 344.9097
~~~

dbenn:
Although a model creation step is not required for FWHM, it could just be simpler from a UI and consistency viewpoint to only show FWHM alongside other error measures in the Models dialog. The only downside is that a model must be created and a period can't be inspected interactively to see FWHM. That could be added later.

dbenn:
Uncertainty output in models dialog now includes std errors and FWHM:
~~~
FWHM lower bound: ...
FWHM upper bound: ...
Standard Error of the Frequency: ...
Standard Error of the Semi-Amplitude: ...
~~~
When harmonics are specified in the model, only FWHM is generated, as per comments in #255.

Issue #255, opened 2022-04-25
DCDFT standard error of the frequency & semi-amplitude


dbenn:
Discussion here gives a way forward for DCDFT period error determination:

  https://www.aavso.org/how-can-we-figure-out-period-error

@BradWalter points out that Grant Foster:

> ...outlines  a few methods of estimating the uncertainty of the estimated frequency of a signal, but recommends the frequency peak FWHM method. This is the one to use since the theoretical methods depend on too many assumptions that are essentially never completely true and often not close to being true.

Foster's methods fall into two categories (page 153 of his _Analyzing Light Curves_ book):

- standard error of the frequency and standard error of the semi-amplitude based upon Fourier model residuals
- FWHM, the period range near the peak over which the periodogram power is at least half its peak value.

Foster also says (page 154 of _Analyzing Light Curves_) that:

> In most cases the possible range estimated by the theoretical formula is too liberal while the range estimated by FWHM is too conservative.

Both methods can implemented in VStar. The standard errors would be presented by the Models dialog, as is the case already for polynomial fits. The second could be made available for a selected peak period via a `FWHM` button and also in the Models dialog for comparison with the standard error values.

I'm surprised I missed (or, forgot about, if I'm being charitable to myself) this brief treatment of period error determination in Foster. I return to Grant's book often, but apparently not often enough! It's one of those gifts that just keeps giving though. Thanks for pointing out section 7.10 Brad.

Section 7.10 also mentions that Grant's AAVSO R code package includes functions for the standard error and FWHM methods. See `Rcodes` here: https://www.aavso.org/software-directory I've played with some of this code but not all, in particular the `peak1` function that includes R implementations of the standard error and FWHM methods. While the standard error calculations were clear from Grant's book, the FWHM approach needed some thought, but the R code will be all I need.

dbenn:
I'm writing notes about this:

<img width="542" alt="image" src="https://user-images.githubusercontent.com/4115628/165032541-b6ee7b18-f4cd-4093-95fc-b3308870b36c.png">

Questions:

1. Are these standard error formulae valid if harmonics beyond the fundamental period are included in the model, since these will affect the residuals?
2. Further, is the standard error of the frequency also useful in determining the extent to which adding harmonics is ”helpful”?

[...]

PR #262, opened 2022-05-02
255 dcdft standard error of the frequency semi amplitude

dbenn:
Your review would be appreciated Max.


mpyat2:
Hi @dbenn , 

I've tried V965 Cep data (see attachment)
[V965_Cep-V-HJD.tsv.zip](https://github.com/AAVSO/VStar/files/8629267/V965_Cep-V-HJD.tsv.zip)

1) DCDFT with period range (Pmin= 0.0850, Pmax= 0.0851, Resolution= 0.0000001) gave the top period= 0.0850674. I've clicked [Create Model], 1 harmonic and got the following uncertainty:
Standard Error of the Frequency: 0.00000159477
Standard Error of the Semi-Amplitude: 0.00151954642

2) Then I reload the data (Nes star from file...) and made the Fourier model with Analysis->Fourier Model...
The 'Uncertainty' pane of the 'Model Information' dialog showed:

Standard Error of the Frequency: **NaN**
Standard Error of the Semi-Amplitude: 0.00151954642

Is it normal?

mpyat2:
Hi @dbenn ,
I've compared the uncertainty with one from Peranso. Peranso (using DCFT) gives the main frequency = 11.75537 (essentially the same as VStar) with the uncertainty= 0.000055 (much bigger). There are different methods of the calculations of the uncertainty and they may give very different results.

The method of the uncertainty calculation used by Peranso is described in the Peranso help, I've copied the description here, maybe you'll find it useful:
------
Peranso determines the minimum period error or period uncertainty of the dominant period P, by calculating a 1-sigma confidence interval on P, using a method described by Schwarzenberg-Czerny (1). This method is a so-called ‘post-mortem analysis’ and measures the width and heights of peaks/valleys in a Period Window.

In his paper, Schwarzenberg-Czerny points out that most of the classical error estimation methods (some of which are present in other period analysis software) are unreliable. That’s the main reason why these methods are not supported in Peranso, despite their simplicity and speed of calculation.

The period uncertainty method of Schwarzenberg-Czerny requires the so called Mean Noise Power Level (MNPL) in the vicinity of P. The 1-sigma confidence interval on P then is equal to the width of the line at the P – MNPL level.

image
 
The 1-sigma interval corresponds to the width
of the line at level MNPL down from the peak


Finding the MNPL may require some care in practice, as many low-power features appearing in a Period Window, are not due to noise but are window patterns of some periods, and thus should not be taken into account. Peranso automatically calculates an approximated MNPL value to determine the period uncertainty. However, you may decide to estimate the MNPL yourself and to enter its value in the Mean Noise Power Level form. The human eye appears to be a good MNPL estimator: simply look at the Period Window and estimate the mean level of the power spectrum (or equivalent) around P, ignoring all strong lines and their aliases. 

(1) Schwarzenberg-Czerny, A., 1991, Mon. Not. R. astr. Soc., 253, 198-206.

mpyat2:
On the other hand, for this sample data, Peranso's uncertainty estimation of the main frequency looks too conservative. I tried to build phase plots with 1/(Freq - uncertainty) and 1/(Freq + uncertainty) and I saw a big difference, using visual estimation of the best frequency (period) gives a more 'sensible' error than formal estimation. VStar's estimation looks more or less reasonable, probably a bit too optimistic.

dbenn:
Thank you for all the useful input @mpyat2. The implementation here was driven by the forum discussion referred to in #255. My main take away from this was to focus initially on the metrics suggested by Grant Foster, here and in #256.

I definitely take your points. My current thinking however, is to start with the standard errors (#255) and FWHM (#256) and leave open other measures of period uncertainty in future, and indeed to include as many as possible, much in the same way we include AIC, BIC, RMS for polynomial fit models and others are being considered for that (#258).

Also, as Foster also says (page 154 of _Analyzing Light Curves_):

> In most cases the possible range estimated by the theoretical formula is too liberal while the range estimated by FWHM is too conservative.

So, I'm hoping we'll get a counterbalanced view of period error initially this way. Then in a future release, we add more, whereas right now, we have none.

Does that seem reasonable?

mpyat2:
Yes, it is reasonable. For now, it is a good starting point.

BradWalter:
I agree supplying both the analytical Standard error of the peak frequency and the FWHM. The standard error almost always underestimates and the FWHM almost always overestimates so you get estimates of lower and upper boundary. 

 

Please note at least for FWHM (not sure about Standard it is important that frequency spacing not be larger than the recommended Frequency grid spacing in eq 7.9 when estimating the FWHM since observations are not evenly spaced in time. The width changes as you increase spacing between frequency samples Particularly above the critical frequency spacing (Δν = 1/(Nτ) if samples are equally spaced in time). 

 

Also It must be clearly understood that the “A” in the standard error  formula is what foster calls the semi-amplitude (absolute value of difference between average and a peak) which is what engineers call Peak amplitude. Amplitude, as used by Foster is what engineers call peak-to-peak (positive direction peak – negative direction peak).  

 

Brad

dbenn:
Thanks for the additional info @BradWalter. Yes, "A" is the semi-amplitude taken from the DCDFT result in the implementation.

dbenn:
@mpyat2 The bug you found arises from the fact that semi-amplitude (A) is used in the calculation of standard error of the frequency and is taken from the DCDFT result for the selected period. When a Fourier model is created independent of a DCDFT, semi-amplitude is not available. So, what I suggest is that for Fourier models, standard error of the frequency is not reported. See next commit.

mpyat2 (approved):

Issue #261, opened 2022-05-02
"Create B-V series" plug-in does not utilize all possible B, V pairs


mpyat2:
See an example:
1) Load data for MQ Cas: Johnson V, Johnson B observations, All times, Observer = PMAK
![image](https://user-images.githubusercontent.com/44919945/166223727-73632f62-d34b-4208-8f32-797d1a7e3477.png)
2) Create B-V series (tolerance=1): only 4 pairs will be used, and 4 other pairs will be ignored. Note that these are DSLR observations (sensor with Bayer pattern) so V and B observations have the same time.
![image](https://user-images.githubusercontent.com/44919945/166223897-550071d9-556e-45ff-bd0c-bc38bed06b5e.png)

If we choose tolerance=0.01, only the first two pairs will be processed.

PR #263, opened 2022-05-02
#261: fixed bug in which we were advancing too far, skipping a pair, …

dbenn:
…when a B,V pair were not within time tolerance

Please review the code and solution, thanks Max.

Plugin test still passes with the change and this seems to fix the bug reported.

### DIFF
--- a/plugin/src/org/aavso/tools/vstar/external/plugin/BMinusVObservationSource.java
+++ b/plugin/src/org/aavso/tools/vstar/external/plugin/BMinusVObservationSource.java
@@ -284,18 +284,16 @@ public class BMinusVObservationSource extends ObservationSourcePluginBase {
 								// code and within the time tolerance requested.
 								bAndVObsSubset.add(first);
 								bAndVObsSubset.add(second);
+								// Move onto the next pair (skip over the current pair).
+								i += 2;
+								continue;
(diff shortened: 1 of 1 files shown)


mpyat2 (approved):
Thank you, @david, it works now as expected so I approve it.
However, as I see it, the approach used in this plug-in implies that there are close B, V pairs separated by long intervals. There are, however, time series where B, V observations are almost equidistant on both sides (imagine that someone makes alternated B, V observations with filter wheel). In this case the simplest linear interpolation between adjacent points would make sense. Also, in the future, we could think about generalizing the plugin to an arbitrary pair of filters.

dbenn:
> Thank you, @david, it works now as expected so I approve it.
> However, as I see it, the approach used in this plug-in implies that there are close B, V pairs separated by long intervals. There are, however, time series where B, V observations are almost equidistant on both sides (imagine that someone makes alternated B, V observations with filter wheel). In this case the simplest linear interpolation between adjacent points would make sense. Also, in the future, we could think about generalizing the plugin to an arbitrary pair of filters.

Thanks and agreed @mpyat2. I've created #264 for the second issue which I've been meaning to for awhile. Would you like to create an issue for the interpolation aspect?

PR #271, opened 2022-05-14
256 dcdft fwhm as a measure of period error

dbenn:
Okay, I've implemented FWHM @mpyat2 if you would not mind reviewing it.

Uncertainty results appear in the models dialog along with std errors now.

I have not yet added a button to show FWHM independent of model creation even though it is not tied to that, because it seems most useful in combination with the standard error of the frequency which is what the original context was.

### DIFF
--- a/src/org/aavso/tools/vstar/util/model/PeriodAnalysisDerivedMultiPeriodicModel.java
+++ b/src/org/aavso/tools/vstar/util/model/PeriodAnalysisDerivedMultiPeriodicModel.java
@@ -158,18 +158,17 @@ public class PeriodAnalysisDerivedMultiPeriodicModel implements IModel {
 		String strRepr = functionStrMap.get(LocaleProps.get("MODEL_INFO_UNCERTAINTY"));
 
 		if (strRepr == null) {
-			if (harmonics.size() == 1) {
-				// Full Width Half Maximum
-				
-				try {
-					Pair<Double, Double> fwhm = fwhm();
-					strRepr = "FWHM lower bound: " + NumericPrecisionPrefs.formatOther(fwhm.first) + "\n";
-					strRepr += "FWHM upper bound: " + NumericPrecisionPrefs.formatOther(fwhm.second) + "\n";
-				} catch (AlgorithmError e) {
-					// don't report this uncertainty
-					strRepr = "";
-				}
+			// Full Width Half Maximum
+			try {
+				Pair<Double, Double> fwhm = fwhm();
+				strRepr = "FWHM lower bound: " + NumericPrecisionPrefs.formatOther(fwhm.first) + "\n";
+				strRepr += "FWHM upper bound: " + NumericPrecisionPrefs.formatOther(fwhm.second) + "\n";
+			} catch (AlgorithmError e) {
+				// don't report this uncertainty
+				strRepr = "";
+			}
 
+			if (harmonics.size() == 1) {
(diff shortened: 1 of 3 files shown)


mpyat2 (approved):
Hi @dbenn , I approve the request in general, however, I think it would be nice to make some small changes:
1) Explicitly indicate that the FWHM uncertainty is for frequency.
2) Add also uncertainty for the period = 1/f.
3) For the standard error of the semi-amplitude: the model function even in the simplest case consists of sin and cos parts having their own amplitudes that differ from the resulting amplitude. Is it possible to indicate the resulting amplitude to which the error refers?

dbenn:
@mpyat2, here is one way to address your suggestions:
~~~
For frequency: 0.060918525, period: 16.415367836, amplitude: 0.397430612: 
  FWHM lower bound: 0.060650162
  FWHM upper bound: 0.061455252
  Standard Error of the Frequency: 0.000016567
  Standard Error of the Semi-Amplitude: 0.011405734
~~~
The FWHM and standard error metrics apply to a frequency, period, amplitude triple, or put another way: a row in the top-hits table. Do you agree?

mpyat2:
Ok, probably the line 'For frequency 0.060918525, period 16.415367836, power 262.29192674, semi-amplitude 0.397430612: ' is sufficient,  I would add 'for frequency' near FWHM to avoid possible ambiguity.

dbenn:
Something like this, also spreading things out a bit?
~~~
For frequency 0.060918525, period 16.415367836, power 262.29192674, semi-amplitude 0.397430612:

  Standard Error of the Frequency: 0.000016567

  Standard Error of the Semi-Amplitude: 0.011405734

  FWHM for frequency:
    Lower bound: 0.060650162
    Upper bound: 0.061455252
~~~
~~~

mpyat2:
Very good!

dbenn:
Feel free to review the changes to src/org/aavso/tools/vstar/util/model/PeriodAnalysisDerivedMultiPeriodicModel.java @mpyat2.

dbenn:
@mpyat2, I could also add the FWHM +/- delta, making it easier to compare with standard error of the frequency. There's a couple of ways this could be done:

- `(upper-lower)/2 = (0.061455252-0.060650162)/2 = 0.0004025450` so:
  - `+/- 0.0004025450` or simply `0.0004025450` 

- Or, next to lower and upper bound, `frequency-lower = 0.060918525-0.060650162` and `upper-frequency` = `0.061455252-0.060918525`:
  - `Lower bound: 0.060650162 (-0.000268363)`
  - `Upper bound: 0.061455252 (+0.000536727)`

As a sanity check, not that `(0.000536727+0.000268363)/2 = 0.0004025450` which is the same as `+/- 0.0004025450`.

To reduce complexity, I tend to think simply `0.0004025450`, e.g.
~~~
FWHM for frequency:
    Lower bound: 0.060650162
    Upper bound: 0.061455252
    Resulting error: 0.0004025450
~~~
Compare this against:
~~~
Standard Error of the Frequency: 0.000016567
~~~
What do you think?

mpyat2:
Looks good, @dbenn

dbenn:
@mpyat2 if you can review the latest commits, I think we're basically done with this issue

dbenn:
Try the latest commit @mpyat2

mpyat2 (approved):
All is OK!

dbenn:
Thanks @mpyat2!

PR #279, opened 2022-05-29
278 broken fourier model cannot find specified frequency in dcdft result error

dbenn:
Thanks @mpyat2 in advance for your review.


mpyat2 (approved):
Thank you, @dbenn , it seems it is fixed now

dbenn:
Thanks for spotting it @mpyat2

Issue #294, opened 2022-06-28
Creating Fourier models sometimes fails with "cannot find specified frequency" error


dbenn:
Creating Fourier models sometimes fails with "cannot find specified frequency" error:

<img width="816" alt="cannot find freq" src="https://user-images.githubusercontent.com/4115628/176168620-67cd2cbe-4711-4afd-ab75-28957d0e2fbf.png">

dbenn:
Another approach would be to take a different approach to looking up the necessary data in period search results that leads to this bug (e.g. result line hashcodes).

dbenn:
Matthias Kolb emailed me with another example of this error.

dbenn:
I'm leaning towards the tolerance based approach, e.g. 1e-5 vs 1e-9. Also, catching the exception in the model class's `toUncertaintyString()` method for std error calcs not just for FWHM, since this is where the failure is currently being exposed.

dbenn:
The problem is however, that it may not matter how much the tolerance is tweaked. It's no guarantee of always finding the right value. Better would be storing the indices in the top hits and full result tables in the model or algorithm object. Or, the hashcode of values could be used. Shouldn't the hashcode of the same interned double value in each table be the same? Test that.

dbenn:
It gets weirder. The output of 3 matched full result and top hits table row pairs:
~~~
5.667367, 5.667360
5.667367, 5.667360
5.667367, 5.667360
~~~
~~~
5.671000, 5.670992
5.671000, 5.670992
5.671000, 5.670992
~~~
5.674114, 5.674113
5.674114, 5.674113
5.674114, 5.674113 
~~~

dbenn:
It gets weirder. The output of 3 matched full result and top hits table row pairs:
~~~
5.667367, 5.667360
5.667367, 5.667360
5.667367, 5.667360
~~~
~~~
5.671000, 5.670992
5.671000, 5.670992
5.671000, 5.670992
~~~
5.674114, 5.674113
5.674114, 5.674113
5.674114, 5.674113 
~~~

dbenn:
The internal handling will be simplified if we know that a frequency/period selection for model creation is a top-hit. The simplest way to do this is to remove the model button from the data table pane, leaving it only on the top hits data pane.

dbenn:
Using the hashcode of a `PeriodAnalysisDataPoint` to match the top hit against result data appears to work well.

[...]

PR #305, opened 2022-11-15
294 fwhm dialog fix

dbenn:
As per comments in #294 @mpyat2, I think these changes are now all that's needed. Please review when you have the chance.


mpyat2 (approved):
It seems it works now as expected

PR #307, opened 2022-12-27
Bug fix for plugin build on Windows, misc. build improvment

orionlee:
The PR is primarily to fix the plugin build problems on Windows due to incorrect assumption on java files encoding:

```ant
    [javac] C:\dev\VStar\plugin\src\org\aavso\tools\vstar\external\plugin\HipparcosObservationSource.java:42: error: unmappable character (0x9D) for encoding windows-1252
    [javac]  * replacing ΓÇ£76343ΓÇ? with the HIP number of interest.
    [javac]                        ^
    [javac] 1 error
    [javac] 1 warning
```

The PR also:
1. makes the same `encoding="UTF-8"`  on the root `build.xml` , to be defensive.
2. tweaks `.gitignore` so that generated sources for VeLa are ignored, and will not be accidentally committed.

Note: following the similar logic in (2), I believe generated  `src/org/aavso/tools/vstar/ui/resources/RevisionAccessor.java` can be removed from git repository and be added to `.gitignore` as well. It's potentially more intrusive so I shy away from making the change.


dbenn (approved):
These changes look good Sam. Thanks!

dbenn:
Thanks for these first contributions to VStar Sam. It's welcomed. Also, I'll give you branch creation privileges so you don't have to fork the repo.

I will merge this PR with the current changes.

Good question re: `src/org/aavso/tools/vstar/ui/resources/RevisionAccessor.java`. I've never been totally happy with the approach. Your suggestion is reasonable. The key thing is that the `REVISION` and `BUILD_TIME` need to be present in the built classes. Removing the generated file from git and adding it to `.gitignore` would still permit that.

One difference from `VeLa` generated source files is that these are generated both in Eclipse and in Ant, whereas `RevisionAccessor.java` is not, so may yield build problems when debugging in Eclipse, for example.
 
My inclination is to create an issue or at least another PR for this so we can talk about approaches. Would you mind doing that?

orionlee:
> One difference from VeLa generated source files is that these are generated both in Eclipse and in Ant, whereas RevisionAccessor.java is not, so may yield build problems when debugging in Eclipse, for example.

Ah, I was afraid that'd be catches like that. Anyway, it's not too big of a problem.

Issue #310, opened 2023-01-01
VStar cannot load saved 'Residials'


mpyat2:
1) Create any model, go to the 'Residuals' tab and save the dataset using default Type='Download', Delimiter='Comma'
2) Try to open the saved file using the default 'Download or Simple' source. An error "Required index >=0" occurs.
![image](https://user-images.githubusercontent.com/44919945/210166524-8f797d57-25d4-45ba-935e-6c6195eef137.png)

There is something wrong with the standard 'Residuals' series. If you replace 'Residuals' with another name in the file (for example, 'Residuals1') the file gets loaded (as 'Unspecified')

dbenn:
This is reminiscent of another problem we saw awhile ago where the name was missing in obs.

Was the name you added in each row or via a #NAME directive?

mpyat2:
It's not a star name, is a series 'filter' name. I renamed 'Residuals' to Residuals1 by the 'replace all' command in a text editor for each line.

dbenn:
Very odd. Trying to debug.

PR #312, opened 2023-01-05
#310: checking that selected series has observations

dbenn:
Hi @mpyat2. I've committed a simple fix for this. Let me know what you think if you are able to.


mpyat2 (approved):
Everything works so far!

Issue #311, opened 2023-01-01
VStar should quote the "comments" field while saving .csv data


mpyat2:
The "Comments" field may contain commas. Currently, VStar saves comma-separated (.csv) data without taking it into account.
Try to load V405 Dra light curve (Johnson V filter, All times) and save observations into a .csv file. Opening the file via the default "Download or Simple" source gives no data (No observations found). This is because VStar gets confused with commas inside the "Comments" field. The solution is enclosing the "Comments" into double quotes (taking care of possible existing quotes inside the field!)

dbenn:
Good pick-up. Surprised this has not been found before. Thanks Max.

PR #313, opened 2023-01-07
#311: quoting comments field including escaping of quotesw

dbenn:
I have fixed the bug by quoting the comments field and escaping any double quotes in the field by preceding each with a double quote, as per the CSV spec.

I've tested that this works using the V405 Dra data set, by saving and reloading it, and also adding quotes inside a row of data to see if it loads and saving it again to ensure that escaped quotes are preserved.


mpyat2 (approved):
Super! Everything works!

Issue #308, opened 2022-12-27
ANOVA result: getting java.lang.IllegalArgumentException: null


orionlee:
When clicking on ANOVA result, I get "java.lang.IllegalArgumentException: null" error dialog box:

![image](https://user-images.githubusercontent.com/250644/209715542-59120248-423a-4290-aafe-eead3970197f.png)

Clicking on the data points on other tabs, including the plot in Periodogram tab, also results in the same error.

It is not blocking the workflow per se. It's just nuisance.

Data used: TESS [TIC 298734307](https://mast.stsci.edu/portal/Mashup/Clients/Mast/Portal.html?searchQuery=%7B%22service%22%3A%22CAOMDB%22%2C%22inputText%22%3A%22TIC%20298734307%22%2C%22paramsService%22%3A%22Mast.Caom.Cone%22%2C%22title%22%3A%22MAST%3A%20TIC%20298734307%22%2C%22columns%22%3A%22*%22%2C%22caomVersion%22%3Anull%7D), sector 57 2 minute cadence SPOC fits file, SAP_FLUX.

I got the same error from some other TESS fits files and AAVSO database.

**Version:** v2.22.0 and the lastet `master`
The older v2.21.3 does not have the same problem.

dbenn:
@orionlee, I am not yet able to reproduce the problem with the specified dataset and 2.22.0 `master` branch:

<img width="763" alt="image" src="https://user-images.githubusercontent.com/4115628/209757508-9dbd0997-0885-4da0-bfb5-79f6537e2596.png">

Can you tell me what AoV parameters you are using? I've tried a few but for the screenshot:

- min: 0.9
- max: 1
- resolution: 0.0001
- bins: 10

Also feel free to point me to another target with JD range in AAVSO database.

Thanks.

orionlee:
Ah, I think I narrowed down the specific scenario the error happened. It happened in the following workflow, irrespective of the underlying dataset:
1. Run DC DFT on some lightcurve
2. Do not dismiss the DC DFT result dialog
3. Run AoV
4. Click the result of AoV, while DC DFT result dialog is still present.
```

Note: I kept DC DFT dialog because to create a model from AoV period, I needed to create model button from DC DFT result. With the (recent?) addition of "Fourier model..." in Analysis menu, keeping DC DFT result is not strictly necessary anymore, but potentially still to be a nice to have.


[...]

PR #315, opened 2023-01-07
#308: checking whether the coordinate is present in the period analys…

dbenn:
…is data point before trying to update cross-hairs

This seems to solve the problem. Can you confirm @orionlee?


mpyat2 (approved):

dbenn:
Thanks @mpyat2 !

orionlee:
The issue is indeed resolved in the latest `master`.

Issue #314, opened 2023-01-07
TESS/Kepler plugin magnitude conversion potentially skewed by bad data


orionlee:
From: https://www.aavso.org/small-caveat-tesskepler-magnitude-conversion

This note is about a small caveat in magnitudes shown for TESS / Kepler data. Empirically, the possible difference is tiny (< 0.0001 magnitude) in SAP_FLUX, so it's likely to be a non issue for most, if not all, cases. I'm posting this note in case the caveat is an issue for some people.

 
Background: I predominantly use Python / [Lightkurve](https://docs.lightkurve.org/) for analyzing TESS data, and notice some small discrepancy in the magnitude I got from my Lightkurve workflow and VStar results. In the handful of samples I looked into, the largest difference is 0.000099 magnitude.

Both VStar and my Lightkurve workflow derive magnitude by converting normalized flux to magnitude, assuming the median flux is the reference TESS magnitude. The discrepancy turns out to be due to the flux included to calculate median:

[VStar](https://github.com/AAVSO/VStar/blob/0115ad7d12db3c8bf429000e79d88708dd458746/plugin/src/org/aavso/tools/vstar/external/plugin/KeplerFITSObservationSource.java#L326): get the median of all  flux values in the FITS file.
My Lightkurve workflow: instead of using all flux values, it ignores the fluxes of possibly unreliable cadences based on the [quality flags](https://outerspace.stsci.edu/display/TESS/2.0+-+Data+Product+Overview) in the FITS file.
For Lightkurve users, ignoring possibly unreliable cadence is the [default behavior](https://docs.lightkurve.org/reference/api/lightkurve.SearchResult.download.html#lightkurve.SearchResult.download) when reading TESS FITS file (that one can override).
 

As a result, the derived magnitude is slightly different. The max is 0.000099 mag in the handful of samples (all based on SAP_FLUX).

298734307, sector 57:  0.000022
441794509, sector 57:  0.000003
261136679, sector 1:  0.000015
198537349, sector 23:  0.000055
323968843, sector 39:  0.000019
77209028, sector 4:  0.000099
 

PDCSAP_FLUX in practice would have no difference most of the time, because most, if not all, likely unreliable cadences have no values in PDCSAP_FLUX to begin with.

 

[...]

PR #316, opened 2023-01-08
Kepler/TESS and QLP observation sources were updated

mpyat2:
In response to #314: Kepler/TESS and QLP observation sources were updated: reference Kepler or TESS magnitude read from the header is stored and displayed in the new column ("Kepler magnitude" or "TESS magnitude" respectively); the "Quality" field was added to Kepler/TESS (the field already existed in the QLP plug-in).

So now the user may exclude any bad points, check for the median or mean using the "Descriptive statistics by series" plug-in, and shift the baseline with the "Magnitude baseline shifter" (one of the possible approaches).

I've attached a 
[TESS_Kepler_TESSQLP_examples.zip](https://github.com/AAVSO/VStar/files/10369054/TESS_Kepler_TESSQLP_examples.zip)
with samples.

image


dbenn (commented):
Looks great, thanks @mpyat!

Issue #324, opened 2023-01-12
Provide means to know the valid names for columns used in VeLa filter


orionlee:
It'd be great if there is some way to know the names for the columns to be used in a VeLa filter. 

Right now one can find out some of them from the examples in [user manual](https://github.com/AAVSO/VStar/raw/master/doc/user_manual/VStarUserManual.pdf) or [wiki](https://github.com/AAVSO/VStar/wiki/VeLa), but one cannot know easily if a column can be used in VeLa filter, and if so, the valid spelling.  E.g., can one filter on calendar date?

It might be a headache to document them, given they are source specific. 

One possibility is that VStar UI provides a list of valid names for the current source.

dbenn:
The list in the dialog UI was not being populated early. Fixed that.

dbenn:
There are names appearing by default that should not be since each source is different, e.g. obscode will appear in AID data but not in TESS.

dbenn:
This requires some thought to implement properly. Working through a few approaches. In principle should be simple, but less so than I'd like.

orionlee:
@dbenn Maybe I misunderstand how the issue is addressed. But when I run the latest VStar on `master` (d2d2317d) , and open a TESS SPOC or QLP fits file, and use View > VeLa Filter

- `quality` (specific to TESS) is not in the Observation Properties dropdown
- some of the options in the dropdown are not applicable, e.g., obscode, shortband, series

![image](https://user-images.githubusercontent.com/250644/215343991-34dd6b49-415b-425c-b209-22b1cf206b88.png)

In contrast, the drop down for regular expression filter reflects the actual available columns:

![image](https://user-images.githubusercontent.com/250644/215344348-7519326c-41d7-4b66-8544-be1cb97fe396.png)

dbenn:
@orionlee: as per the new pull request, I've fixed the arbitrary column names case. I've also removed obscode as a special case. There are other names than columns no matter what the data source, e.g. as a convenience for time (t, jd, ...), magnitude (mag), and band (shortband, e.g. V vs Johnson V), series (can be the same as band, but not always, specifically when a user defined series is created, e.g. from a filter). Some names have underscores added for VeLa but are mapped back to the underlying name.

PR #330, opened 2023-01-31
#324: fixed arbitrary column source names getter; removed obscode as …

dbenn:
…as special case

@orionlee, I've properly accounted for the arbitrary column names case and removed obscode as a special case

### DIFF
--- a/src/org/aavso/tools/vstar/ui/model/list/ArbitraryFormatRawDataColumnInfoSource.java
+++ b/src/org/aavso/tools/vstar/ui/model/list/ArbitraryFormatRawDataColumnInfoSource.java
@@ -17,8 +17,10 @@
  */
 package org.aavso.tools.vstar.ui.model.list;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.aavso.tools.vstar.data.ValidObservation;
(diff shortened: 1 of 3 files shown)


orionlee:
I tried the code in the PR, and verified that it worked for QLP source: quality is now included in the dropdown!

src/org/aavso/tools/vstar/ui/model/list/ArbitraryFormatRawDataColumnInfoSource.java
```diff
@@ -69,7 +71,9 @@ public int getColumnCount() {
 
 	@Override
 	public Collection<String> getColumnNames() {
-		return COLUMN_NAMES.keySet();
+		List<String> names = new ArrayList<String>(COLUMN_NAMES.keySet());
+		names.addAll(ValidObservation.getDetailTitles().keySet());
```
orionlee, line 75: I don't know the codes well so this might not be applicable.
Logically, the dropdown in Observations tab > Regular Expression search is similar to the VeLa filter dropdown here. 

But it seems to be based on different logic: I did a search on `ValidObservation.getDetailTitles()` and couldn't find it elsewhere. 

If they are indeed based on different logic, consider to consolidate them.

dbenn: I'm occupied with a few things but will return to this in a day or so.

mpyat2 (approved):

src/org/aavso/tools/vstar/ui/model/list/ArbitraryFormatRawDataColumnInfoSource.java
dbenn: @orionlee, thanks for the comment. The change led to the VeLa filter and the observation list regex search using the same approach, and not `ValidObservation.getDetailTitles()` (on the branch).

PR #338, opened 2023-03-25
Issue 337 gaia plug in enhancement

mpyat2:
Hi @dbenn ,
Here are severely reworked GAIADR2XformObSource (retrieving Gaia data from the Internet) and the new GAIADR2XformFileObSource, which reads from a local file. Both extend the base class GaiaObSourceBase, which is moved to a separate shared jar, placed in vstar_plugin_libs. PluginManager refused to work with them (it could not create an instance of the Gaia class because it did not retrieve the shared jar from a remote location). So I fixed the PluginManager too.
I also made some additional PluginManager fixes and refactoring. Although it was not directly related to the Gaia plug-ins, as far as the PluginManager needed to be fixed anyway, I decided to make more fixes. One important fix was the new areURLReferentsEqual() method. The old one ALWAYS gave FALSE. The new one, I hope, works correctly.

### DIFF
--- a/src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
+++ b/src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
@@ -438,6 +438,7 @@ public class PluginManager {
 				// Separate jar file name from dependent libraries, if any
 				// exist.
 				String[] fields = line.split("\\s*=>\\s*");
+				//System.out.println(fields.length + " " + line);
 
 				if (fields.length != 0) {
 					// Load plugin, store mappings from jar name to plugin
@@ -455,15 +456,27 @@ public class PluginManager {
 						pluginJarFileName = pluginJarFileName.substring(1);
 						// }
 					}
-
+					
 					URL pluginUrl = new URL(pluginBaseURLStr + "/"
 							+ pluginJarFileName);
 					String className = pluginJarFileName.replace(".jar", "");
-
+					
+					//System.out.println(pluginUrl);
+					
+					List<URL> depLibs = new ArrayList<URL>();
+					if (fields.length == 2) {
+						for (String libFileStr : fields[1].split("\\s*,\\s*")) {
+							String libJarFileName = libFileStr;
+							depLibs.add(new URL(libBaseURLStr + "/"	+ libJarFileName));							
+						}
+					}
+					
 					//Max: get class loader with plugin and close it at the end!
                     // https://docs.oracle.com/javase/8/docs/technotes/guides/net/ClassLoader.html
 					Pair<IPlugin, URLClassLoader> plugin_cl = createObjectFromJarURL(pluginUrl,
-							className);
+							className, depLibs);
+					//System.out.println("Remote plug-in info: " + plugin_cl.first.getDescription());
+					//System.out.println();					
 					try {
 						remoteDescriptions.put(plugin_cl.first.getDescription(),
 								pluginJarFileName);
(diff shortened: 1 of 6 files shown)


src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
```diff
@@ -754,7 +759,12 @@ public void deletePlugin(String description) {
 					// vagaries of file systems or the possibility of
 					// concurrent deletion.
 					if (libJarPath.exists()) {
-						// if (libJarPath.delete()) {
+						// Max: This fragment was commented out by David(?)
+						// Even being commented out, it contained an error in the if-condition:
+						// logical '!' was missing!
```
dbenn, line 764: agreed

src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
```diff
@@ -754,7 +759,12 @@ public void deletePlugin(String description) {
 					// vagaries of file systems or the possibility of
 					// concurrent deletion.
 					if (libJarPath.exists()) {
-						// if (libJarPath.delete()) {
+						// Max: This fragment was commented out by David(?)
```
dbenn, line 762: I'm wondering whether I commented it out because of a problem with deletion of files from WebStart launched VStar. Would need to search history. So long ago it may have been svn not git.

src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
```diff
@@ -754,7 +759,12 @@ public void deletePlugin(String description) {
...
+						// Even being commented out, it contained an error in the if-condition:
+						// logical '!' was missing!
+						// I have not uncommented it but fixed the error.
+						// So, actually, we never delete the libraries. 
+						// if (!libJarPath.delete()) {
```
dbenn, line 767: That's right. See also my comment above. Arguably we could allow this deletion to occur now that we don't use WebStart, if that is indeed the reason. Speculation at the moment.

src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
```diff
@@ -881,30 +891,43 @@ private void copy(InputStream in, File file) throws IOException {
...
+			IPlugin plugin = (IPlugin) clazz.newInstance();
+			return plugin.getDescription();
+		} finally {
+			loader.close();
+		}
 	}
 
 	private boolean areURLReferentsEqual(URL url1, URL url2) throws IOException {
-		return getBytesFromURL(url1).equals(getBytesFromURL(url2));
+		// Max: arrays retrieved by getBytesFromURL will never be equal!
```
dbenn, line 911: True! Good pick up. There should really be no `equals()` method for arrays in Java since this and `==` are identical.

src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
```diff
@@ -913,4 +936,22 @@ private List<byte[]> getBytesFromURL(URL url) throws IOException {
 
 		return byteList;
 	}
+
+//	private List<byte[]> getBytesFromURL(URL url) throws IOException {
```
dbenn, line 940: Feel free to just delete this and use this name for your new implementation Max.

dbenn (commented):
Thanks @mpyat2. Minor comments and thoughts have been added as review comments. Have a look and see what you think. Let me know if there are further changes coming from these. I can review any further changes or just approve and merge.

mpyat2:
Hi @dbenn ,
I made fixes according to your remarks and also fixed (I hope) dependent lib ref counting plus MessageBoxes calling via InvokeLater (they always recommend not to call interface-related methods in the background thread, I wonder why we still have not had any problems with this. Maybe MessageBoxes is something special). 
I also added Cliff to the reviewers because I made severe changes to his work.
Well, there was too much side work with the PluginManager here, I should stop it and create a ticket if something more is required. Let's see.

dbenn (approved):

Issue #342, opened 2023-03-27
Kepler/TESS, QLP, and Lightkurve plug-ins: make them derived from TESSObSourceBase to reduce code duplication


mpyat2:
The plug-ins Kepler/TESS, QLP, and Lightkurve are pretty similar; we can reduce code duplication by making a base class for them.

PR #345, opened 2023-03-31
TESS-related plug-ins optimization

mpyat2:
Issue #342: KEPLER/TESS, QLP, LightKurve plug-ins now derive their ObservationRetriever's from TESSObservationRetrieverBase. 
A new observation detail -- "Flux" -- was added.
FITS rows having bad flux/flux_error are now collected to the invalid observations list.


dbenn (approved):
Great stuff @mpyat2! Thanks!

PR #348, opened 2023-04-09
339 user created series do not persist across additive loads

dbenn:
@mpyat2 There's a number of things going on in this pull request and the changes are deep in the guts of VStar, so feel free to ask questions. A review is the key thing. Thanks.

### DIFF
--- a/src/org/aavso/tools/vstar/data/ValidObservation.java
+++ b/src/org/aavso/tools/vstar/data/ValidObservation.java
@@ -1280,11 +1280,6 @@ public class ValidObservation extends Observation {
 		return buf.toString();
 	}
 
-	/*
-	 * (non-Javadoc)
-	 * 
-	 * @see java.lang.Object#hashCode()
-	 */
 	@Override
 	public int hashCode() {
 		final int prime = 31;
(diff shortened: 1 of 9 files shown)


mpyat2:
I noticed a strange behaviour:
1. New Star from AAVSO...
   a) RR LYR, from 2459103 to 2459104, Johnson V
   b) View->Filter from plot
   c) View->Create series...->Filtered (select some distinct color, i.e. red)
   d) Tools->Heliocentric JD converter (only original series will be converted, this is an old 'feature' (or bug?))
2.  New Star from AAVSO...
   a) RR LYR, from 2459096 to 2459097, Johnson V, Add to current = yes

Now zoom the first part, say, from 2459103.35 to 2459103.39. You will see doubled V observations for both JD and HJD.
    
image

This strange duplication existed before and is not directly connected with the current changes, but I only noticed it now.
The duplication appears only if a "New series" is created.

dbenn:
I can reproduce this on the current branch but not with 2.22.0.

Step 1. b) is not needed BTW. You can just create a series via View -> Create Series using Johnson V (vs Filtered).

The duplicate observations must be entering through the new-series-in-presence-of-additive-load logic.

dbenn:
Okay, fixed! Geez! See last commit, but in the end, all the craziness I added/changed in `Mediator` was not necessary. Which is good, because it felt wrong, had a bad smell. Seemed like an anti-pattern. Instead, a 2 lines change in `AbstractObservationRetriever.categoriseValidObservation()` was all that was required to properly handle user-defined series. Please check again @mpyat2. Note that this last commit does not address the user-defined-series-does-not-get-HJD-converted bug you noted above. We can either do that here or in another issue.

mpyat2 (approved):
Looks good, @dbenn

Issue #350, opened 2023-04-16
ASAS-SN Sky Patrol, Version 2.0: the plug-in must be enchanced


mpyat2:
See https://arxiv.org/abs/2304.03791
and http://asas-sn.ifa.hawaii.edu/skypatrol

PR #352, opened 2023-04-27
Issue 350 asassn v2

mpyat2:
ASAS-SN introduced SkyPatrol 2.0 [http://asas-sn.ifa.hawaii.edu/skypatrol/] with the new data format. The plugin has been modified to support it along with all the old formats. See #350


dbenn (approved):
Thanks Max

PR #355, opened 2023-05-05
Issue 351 ztf file plugin

mpyat2:
I added a new ZTF file-based plug-in. It shares code with the ZTF-internet-source plug-in (ZTFObSource) so the new shared lib was created also.
The new "Copy ZTF URL to clipboard" button was added to the ZTF Photometry dialog (and the similar one to the GAIADR2XformObSource)


dbenn (approved):

dbenn:
Thanks @mpyat2 !

PR #358, opened 2023-06-01
331 mac os x missing files in file open dialog problem persists

dbenn:
Hi @mpyat2 

I have finally fixed this. I will get some Mac people to test it but feel free to have a look over this. I can just merge it I think.

David


mpyat2 (approved):
Hi @dbenn 
I cannot test this unfortunately so I simply approve...

Issue #373, opened 2023-10-06
PluginManager cannot correctly check for local and remote plug-in equality


mpyat2:
It is supposed that PlugInManager enables the Update button for a particular plug-in only if the local and remote copies are non-equal.
The problem is in getBytesFromURL() method: it was supposed that stream.available() returns the stream length. This is actually not the case: "Note that while some implementations of InputStream will return the total number of bytes in the stream, many will not. It is never correct to use the return value of this method to allocate a buffer intended to hold all data in this stream" (see the Java documentation).

It **does** work if the remote path is a local directory and **sometimes** works for Web (for instance for my test website http://mpta.atwebpages.com/vstar/plugins/ ; even for this site sometimes it returns 0) but does NOT work for www.aavso.org.

I'm sure we can fix the issue (it is not critical anyway, rather 'cosmetic'). But, @dbenn , do we really need to load the plug-ins from a remote location? Probably, for Linux users, it would be helpful. If yes, we could simplify PluginManages, leaving a single [Install/Update] button instead of the two ones. Anyway, we rely on the plug-in equality solely without checking what is newer.

dbenn:
@mpyat2, does "do we really need to load the plug-ins from a remote location" relate to #192?

As you go on to say:
> ...we could simplify PluginManages, leaving a single [Install/Update] button instead of the two
we could just unconditionally install whether or not it's changed, as you say.

PR #378, opened 2023-10-29
Fixed areURLReferentsEqual; PluginManager dialog tuning

mpyat2:
Fixed areURLReferentsEqual; in the PluginManager dialog, an item is grayed if local and remote plug-ins differ or if the remote plug-in does not exist.

### DIFF
--- a/src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
+++ b/src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
@@ -983,19 +983,16 @@ public class PluginManager {
 		List<Byte> byteList = new ArrayList<Byte>();
 		InputStream stream = url.openStream();
 		try {
-			int len = stream.available();
-			while (len > 0) {
-				byte[] bytes = new byte[len];
-				stream.read(bytes, 0, len);
-				for (byte b : bytes) {
-					byteList.add(b);
+			byte[] buf = new byte[4096];
+			int len;
+			while ((len = stream.read(buf)) > 0) {
+				for (int i = 0; i < len; i++) {
+					byteList.add(buf[i]);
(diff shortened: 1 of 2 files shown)


mpyat2:
Issue #373 fix

src/org/aavso/tools/vstar/ui/dialog/plugin/manager/PluginManager.java
```diff
@@ -983,19 +983,16 @@ private List<Byte> getBytesFromURL(URL url) throws IOException {
 		List<Byte> byteList = new ArrayList<Byte>();
 		InputStream stream = url.openStream();
 		try {
-			int len = stream.available();
-			while (len > 0) {
-				byte[] bytes = new byte[len];
-				stream.read(bytes, 0, len);
-				for (byte b : bytes) {
-					byteList.add(b);
+			byte[] buf = new byte[4096];
```
dbenn, line 986: Do we need to be careful about the use of 4096 here @mpyat2 ?

mpyat2: Hi @dbenn , well, actually, this is a somewhat modified 'copy' method (see line 945) that has worked for years. What could be wrong with a 4096 buffer?

dbenn: @mpyat2, you're right of course. I'm thinking like a C programmer! Progress on the road to senility? :) I'm happy from https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read-byte:A- that "The number of bytes read is, at most, equal to the length of b."

dbenn (approved):
Looks good. Thanks @mpyat2!

PR #380, opened 2023-11-11
Plugin manager help functionality

mpyat2:
Hi @dbenn ,

Please take a look at this if you have time. In addition to the 'Help' functionality implemented in the plug-in manager dialog, I've added 'Help' button to some plug-in dialogs (New star from AAVSOnet/APASS, Create B-B series, Gaia, ZTF, New star from VeLa, Fourier Model (Period count/ periods dialogs), Heliocentric JD converter.

Ironically, with this approach, the plug-ins have better help support than the main program. Probably we could think about a better help subsystem (in the future).

Cheers,
Max


src/org/aavso/tools/vstar/plugin/IPlugin.java
```diff
@@ -52,6 +52,11 @@ public interface IPlugin {
 	 */
 	abstract public String getDescription();
 
+	/**
+	 * Get a document file name for this plugin.
+	 */
```
dbenn, line 57: It would be worth adding that this document can be a full URL or a file name that assumes the plugin doc dir as  a base URL. This could help plugin developers.

dbenn (commented):
This looks good @mpyat2.

One other thing to consider is adding a Help button `AdditiveLoadFileOrUrlChooser`. The `plugins` map in this class can be used to invoke `getDocName()` on the selected plugin's object.

dbenn:
I agree that the plugin context-sensitive help is great and would be good to do for the tool itself. The 130+ page user manual was a **_big_** undertaking at the time, especially in my spare time, and at least the hard part is done (the content). If I was doing it again I'd put the manual on a wiki, which would make pointing to sub-sections easier; but this was pre-GitHub times.

mpyat2:
Hi @dbenn , yes, if the manual is converted to separate pages, they could be used as help topics.
Meanwhile, I've added a button to the chooser (file open dialog). I've also created a stub for the "AAVSO download and simple formats" source Wiki page (the internal plug-in). Slightly lately, I will push the changes.
image

mpyat2:
OK, I pushed the changes. Now, file chooser contains a help [?] button that invokes the plugin's help.

dbenn:
Thanks @mpyat2. Anything else to be done or happy for me to merge?

mpyat2:
Hi @dbenn , 
please merge it!

Issue #382, opened 2023-11-17
The Harmonics dialog is displayed twice


mpyat2:
1. Do "DC DFT"
2. Select a point
3. Press the "Find Harmonics" button.

The "Harmonics" dialog is displayed. Press Dismiss. The same dialog will be displayed one more time.

mpyat2:
see also https://github.com/AAVSO/VStar/issues/61#issuecomment-720451600

mpyat2:
see the branch '[period-analysis-messages](https://github.com/AAVSO/VStar/tree/period-analysis-messages)'

dbenn:
I have definitely seen this bug before. Thanks.

mpyat2:
I address this bug in #383, too.

Issue #381, opened 2023-11-12
Period analysis plugin dialogs share period selection notification messages.


mpyat2:
If the user selects a new period in any period analysis dialog (DC DFT, WWZ), the notification message propagates through all open period analysis dialogs. To reproduce: do DC DFT, then do not close the dialog and do WWZ. Then, try to select a period in the table or in the plot. The selection in the other dialog will change.
It's confusing.

Possible solutions:
1) add a dialog identifier to the message and process only the messages that relate to the specific dialog instance
2) a simpler solution: allow one and only period analysis dialog to be opened at the same time: 
      add `private static PeriodAnalysisDialogBase prevInstance` to PeriodAnalysisDialogBase
      add the following code to the constructor at the beginning:
```
		if (prevInstance != null) {
			if (prevInstance.isVisible()) {
				prevInstance.setVisible(false);
				prevInstance.cleanup();
				prevInstance.dispose();
			}
		}
		prevInstance = this;
```

dbenn:
I have not hit this problem because I don't think I've ever had two PA dialogs open at the same time.

So, solution 2 would close the first dialog, right? That may or may not be what the user expects I suppose. A variant would be to open a message dialog that says: only one PA at a time permitted.

Solution 1 is a good idea. It's not normally the way the Observer pattern works of course but it is somewhat like protocols such as MODBUS.

There may be another solution relating to message listener lifetimes and window focus. That needs a closer look though.

orionlee:
One data point: When I use VStar (albeit only occasionally), I often run a DC DFT, followed by an AoV, and would keep the 2 dialogs open:

![image](https://github.com/AAVSO/VStar/assets/250644/db9819cf-3c9a-4423-aa26-7d97607944e0)

mpyat2:
... which makes the 2nd totalitarian-style solution inappropriate.

orionlee:
BTW, the issue seems to be related to the old #308 .

mpyat2:
A side finding: sometimes, too many "PeriodAnalysisSelectionMessage" messages are generated.
[...]

Issue #61, opened 2020-09-26
Harmonics search should give a more helpful message if frequency doesn't exist in data (e.g. a CLEANest value added to top-hits)


dbenn:
Brad W reports:

> If I select a tophit and click on the button in some cases It gives harmonics other than the 1st however, if I click on the harmonics in the list that comes up. Nothing seems to happen in some cases but for other harmonics an error dialog opens up Java.lang.IllegalArgumentException: Unknown frequency. Does that mean this frequency is not really a harmonic but the ones for which nothing happens are harmonics. 
>  
> If I apply find harmonics to a cleanest output frequency it only shows the fundamental even if harmonics are listed for the frequency to which cleanest was applied. Also if I click on the 1st Harmonic frequency (the only frequency listed) in the find harmonics output for a CLEANest result, I always get the above Java.lang Unknown frequency error.

dbenn:
Additional detail from Brad:

> I get the same error for harmonics of the original DCDFT generated frequencies in the tophits. For example the attached shows 19 harmonics of this DCDFT tophit frequency (not CLEANest) but all except the fundamental throw this error. That makes me question whether these are actually harmonics rather than a bug in the harmonic function. Is it possible I get the error because there are result in the data file that are not included in the tophits selected from the data file and when you click on the frequency in the list located in the data file by the harmonic tool it checks whether it is also a tophit as a means of checking significance?  
>  
> The way I am doing my search for signal in this very complicated light curve is
> DCDFT the original LC pick the most powerful freq,
> CLEANest that Frequency & check the tophit freq generated by DCDFT (Not CLEANest) for harmonics
> Create a model with the CLEANest frequency
> DCDFT residuals Select the highest frequency, CLEANest that frequency  & check the tophit freq generated by DCDFT (Not CLEANest) for harmonics
> Return to the original LC data Run DCDFT
> CLEANest all frequencies discovered to that point all at once update (manually kept) list of discovered frequencies/periods including changes to any previously discovered frequencies cause by added signal.
> Create a model with all freq discovered to that point
> Repeat steps 4-7 iteratively
>  
> My problem is that I don’t trust the harmonics that I get from the harmonic tool because of the error messages generated. I get the error  when I click on all except the 2nd harmonic 0f 0.001234 (the 3rd H throws the error) Right now the harmonics tool has provided harmonics for 6 fundamentals. The number of harmonics shown for each are in the list below.
> 
> Freq | # Harmonics
> -- | --
[...]

PR #383, opened 2023-11-18
Period analysis dialog issues with messages

mpyat2:
This addresses issues #381, #382, and at least partially #61 (IllegalArgumentException while selecting harmonics in HarmonicInfoDialog).

### DIFF
--- a/src/org/aavso/tools/vstar/ui/dialog/period/PeriodAnalysis2DResultDialog.java
+++ b/src/org/aavso/tools/vstar/ui/dialog/period/PeriodAnalysis2DResultDialog.java
@@ -30,7 +30,9 @@ import org.aavso.tools.vstar.plugin.PluginComponentFactory;
 import org.aavso.tools.vstar.plugin.period.PeriodAnalysisComponentFactory;
 import org.aavso.tools.vstar.plugin.period.PeriodAnalysisDialogBase;
 import org.aavso.tools.vstar.ui.NamedComponent;
+import org.aavso.tools.vstar.ui.dialog.DoubleField;
 import org.aavso.tools.vstar.ui.dialog.MessageBox;
+import org.aavso.tools.vstar.ui.dialog.MultiEntryComponentDialog;
 import org.aavso.tools.vstar.ui.mediator.Mediator;
 import org.aavso.tools.vstar.ui.mediator.message.HarmonicSearchResultMessage;
 import org.aavso.tools.vstar.ui.mediator.message.PeriodAnalysisSelectionMessage;
@@ -75,6 +77,11 @@ public class PeriodAnalysis2DResultDialog extends PeriodAnalysisDialogBase {
 	private JTabbedPane tabbedPane;
 	
 	private String findHarmonicsButtonText;
+
+	private static final double FREQUENCY_RELATIVE_TOLERANCE = 1e-3;
+	
+	private double currentTolerance = FREQUENCY_RELATIVE_TOLERANCE;
+	DoubleField toleranceField;
 
 	/**
 	 * Constructor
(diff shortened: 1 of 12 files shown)


mpyat2:
Hi @dbenn , I know you are busy with Auth0; please review this only when you have time after any urgent things.
Probably I will make some additional changes after additional tests, so there is no rush here.

mpyat2:
In the last commit:
1) findHarmonics() tolerance can be tuned by the user.
2) Harmonics are searched within top-hits only, which makes more sense (IMHO - Max)
3) HarmonicInfoDialog: a procedure that shows a selected harmonic was unreasonably complicated. Why should we seek for the y coordinate? Let's display the x-position only.
4) The domain type was ignored in the HarmonicInfoDialog, which made displaying the selected harmonic not work if x of PERIOD type.
5) An algorithm of the hair-cross positioning/showing was changed to make it more intuitive (IMHO - Max)

src/org/aavso/tools/vstar/ui/mediator/message/MessageBase.java
```diff
@@ -22,6 +22,8 @@
  */
 public class MessageBase {
 
+	private String name;
```
dbenn, line 25: I'm wondering about the use of `name` here. The intent is something like "destination" or "recipient" identifier I think. Is that correct?

dbenn (commented):
Hi @mpyat2, everything you're doing looks great and as always, I appreciate it!

I'm just pausing a bit to think through the message notification changes because it's a fairly core part of the way things operate and it's been around a long time.

The use of the identifier (name) would presumably be optional, right?

The fundamental issue seems to relate to the unintended reception of messages. I'm just trying to think of any other way to achieve the same effect that for example, involves the observer not subscribing to particular messages or subscribing to something more specific than is being subscribed to currently.

I can't really flaw the approach you're taking. Just thinking it through Max! :)

I have one suggested change in MessageBase though (see comment in that class).

src/org/aavso/tools/vstar/ui/dialog/period/PeriodAnalysis2DResultDialog.java
```diff
@@ -219,29 +238,57 @@ private JTabbedPane createTabs() {
...
-		List<Harmonic> harmonics = findHarmonics(
-				selectedDataPoint.getFrequency(), data);
+		String chartID = null;
+		Component c = tabbedPane.getSelectedComponent();
+		if (c instanceof PeriodAnalysis2DChartPane) {
+			chartID = ((PeriodAnalysis2DChartPane)c).getChartPaneID();
+		}
+		
+		if (chartID == null) {
+			MessageBox.showMessageDialog("Find Harmonic", "Please select a tab with a chart");
```
dbenn, line 255: I am wondering about this. Why are we not permitting the top-hits (or data) table to be selected? It seems quite intuitive to me for that to be permitted in addition to a plot tab, especially since we are now limiting ourselves to searching top-hits (which is fine) for harmonics.

src/org/aavso/tools/vstar/ui/dialog/period/PeriodAnalysis2DResultDialog.java
```diff
@@ -219,29 +238,57 @@ private JTabbedPane createTabs() {
...
+		List<Harmonic> harmonics = findHarmonics(selectedDataPoint.getFrequency(), data, currentTolerance);
 		HarmonicSearchResultMessage msg = new HarmonicSearchResultMessage(this,
 				harmonics, selectedDataPoint);
+		msg.setName(this.getName());
+		msg.setIDstring(chartID);
 		Mediator.getInstance().getHarmonicSearchNotifier().notifyListeners(msg);
 	}
+	
+	private MultiEntryComponentDialog createParamDialog() {
+		toleranceField = new DoubleField("Relative Frequency Tolerance", 0.0, 1.0, currentTolerance);
```
dbenn, line 275: I like it!

mpyat2:
Hi @dbenn ,

The use of the message identifier is entirely optional. It is something like a tag (label) attached to the message. Currently, it is used in the period analysis dialogs only to distinguish between dialog instances that may coexist if the user leaves the previous dialogs open. I practiced this, and as we can see, I'm not unique. Well, 'name' probably is not the best name for the field (ha-ha, the sentence sounds funny). Maybe 'tag' sounds better?

By the way, to get rid of multiple showing the harmonic dialog, I added a secondary identifier (iDstring) to HarmonicSearchResultMessage. Concerning blocking the harmonics dialog on the tabs with tables: I will try to remove this blocking and implement top-hit/data point selection.

dbenn (approved):

PR #399, opened 2024-01-27
398 load multiple observation files in sequence

dbenn:
Hi @mpyat2, please see what I've written in the issue and let me know if you have questions.

I may add the multiple file feature to other plugins before we close this but for now it's just the infrastructure and TESS-common-code plugins. Indeed, given a Jan 31-ish release deadline due to Auth0 (more like Feb 1 or 2 I'd say), there may not be time for much more.

### DIFF
--- a/plugin/src/org/aavso/tools/vstar/external/lib/TESSObservationRetrieverBase.java
+++ b/plugin/src/org/aavso/tools/vstar/external/lib/TESSObservationRetrieverBase.java
@@ -25,6 +25,7 @@
 
 package org.aavso.tools.vstar.external.lib;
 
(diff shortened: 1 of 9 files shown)


mpyat2 (approved):
Hi @dbenn , it's a great improvement. For big data sequences (like TESS), loading several files at once is much faster than additive loading (when we use additive loading, adding each new bunch of data takes a long time, probably because of relocation of existing data or something similar).
Thank you for the implementation!
In the future, it will be useful to add this functionality to other plugins. However, being under pressure from the upcoming release, it would be wiser to postpone other changes.
It seems it works as expected.

Issue #415, opened 2024-04-24
DownloadFormatObservationSinkPlugin: Problem with "Space" delimiter


mpyat2:
1. New Star from AAVSO Database..., T Umi, Minimum JD= 2460310.5, Maximum JD= 2460425.5
2. Observations, Save -> Type= "Download", Delimiter= "Space", any name (like T_Umi_space_delim.txt)
3. File Open -> Source="Download or Simple"
4. Select the "Observations" tab and look at the "Name" column: identifiers with spaces truncated.

Possible solution:
in ValidObservation.toAAVSOFormatString:
use "quoteForCSVifNeeded" for _all_ text fields (not only for comments).

```
	private String quoteForCSVifNeeded(String field, String delimiter) {
		return field.contains(delimiter) ? quoteForCSV(field) : field;
	}
```

mpyat2:
Hi @dbenn 
Could you please attach a file (**space**-separated) you opened?

mpyat2:
Here is my example 
[T_Umi_space_delim.txt](https://github.com/AAVSO/VStar/files/15108727/T_Umi_space_delim.txt)

And what I see after opening it:
![image](https://github.com/AAVSO/VStar/assets/44919945/1b7eedf3-9e61-4fa3-857b-49f5094c6e09)

dbenn:
> Hi @dbenn
> Could you please attach a file (**space**-separated) you opened?

@mpyat2, here it is; I used the same name for the file :)
[T_Umi_space_delim.txt.csv](https://github.com/AAVSO/VStar/files/15111467/T_Umi_space_delim.txt.csv)

dbenn:
Interesting that you see this problem and I don't on very similar builds. Windows vs Mac? Unlikely unless we're talking about CR+LF etc. `od -c` may provide some insights but I don't see anything immediately obvious in the file.

Very happy for you to go with your proposed solution (your code above) if you want to.

dbenn:
Ah, a key difference is that I saved as CSV and you saved with space delimiters and indeed, you **_did_** say to use `Delimiter= "Space"`. So yes, I can see why you see what you do and that quoted strings would help. So, yes, your proposed code change makes sense. Thanks!

PR #416, opened 2024-04-25
ValidObservation.toSimpleFormatString, toAAVSOFormatString: text fields are double-quoted if they contain the delimiter.

mpyat2:
See #415


dbenn (approved):
Thanks @mpyat2

Issue #414, opened 2024-04-21
FlexibleTextFileFormatObservationSink as a complementary pair to FlexibleTextFileFormatObservationSource


mpyat2:
Create FlexibleTextFileFormatObservationSink  that stores information in the "FlexibleText" format

PR #417, opened 2024-04-25
Issue 414 FlexibleTextFileFormatObservationSink

mpyat2:
See #414


mpyat2:
Hi @dbenn , please do not close this pull request, I found a small glitch and want to fix it first.

mpyat2:
Hi again @dbenn :)

I fixed the glitch, so now the branch is ready for review

plugin/src/org/aavso/tools/vstar/external/plugin/FlexibleTextFileFormatObservationSink.java
```diff
@@ -0,0 +1,219 @@
...
+		buf.append("\n");
+		return buf.toString();
+	}
+	
+	private String quoteForCSV(String field) {
+		return "\"" + field.replace("\"", "\"\"") + "\"";
+	}
+
+	private String quoteForCSVifNeeded(String field, String delimiter) {
+		return field.contains(delimiter) ? quoteForCSV(field) : field;
```
dbenn, line 212: An opportunity for refactoring here since `ValidObservation` has `quoteForCSVifNeeded` and `quoteForCSV` now I think?

dbenn (changes requested):
Hi @mpyat2, thanks for this. One suggested possible refactoring; see comment inline.

mpyat2:
Hi @dbenn , thank you for the suggestion!
You are right; making ValidObservation.quoteForCSVifNeeded(...) _public_ _static_ allows us to eliminate code duplication.
However, now this method is in the parallel universe of the other branch, creating a quantum wormhole between them may lead to something unpredictable (sorry, I'm reading Dan Simmons' Ilium/Olympos cosmic opera now and some nonsense comes into my head :) ). So, it would probably be better to make a new branch after merging this one, and I will refactor the code. What do you think?

dbenn:
Agreed. That sounds good @mpyat2

PR #427, opened 2024-06-08
425 time shifter plugin

dbenn:
Hi @mpyat2  I've converted the mag shifter plugin to include time. The documentation has been rewritten as well.


mpyat2 (approved):
Thank you very much for this very helpful improvement! I especially love the phase plot update: in the JD->BJD and in the JD->HJD plug-ins, this was not implemented and the plot simply changed to the light curve mode. I'll try to implement this later.

dbenn:
Thanks @mpyat2. Yes I was going to mention the phase plot update as something we could also use in these other plugins. :)

PR #428, opened 2024-06-10
424: changed exception handling in text format obs reader

dbenn:
Hi @mpyat2

I'm taking the minimal approach here of propagating the exception (removed the `try`..`catch`, added `throw` clause), allowing it to be caught higher up so that it appears in an error dialog box. There are many things that could be done to improve this code, but that's another issue entirely.

A question is whether or not `ObservationReadError` and `IOException` should be unified, e.g. caught and rethrown as `ObservationReadError`? Happy to make that change although one argument against that is that different exceptions thrown give more information, so leaving as multiple exceptions also makes sense.

Let mew know what you think.


mpyat2:
Hi @dbenn 
I think we can leave ObservationReadError and IOException as they are for now

Issue #394, opened 2024-01-15
Functions with no return value, output  "No value present" in VeLa dialog


dbenn:
Enter `print("1")` into the VeLa dialog and hit the `Run` button.

`1` will be output and after it, `No value present`.

Distinguish between a function that returns no value and a function that does not need to return a value.

PR #434, opened 2024-06-26
#394: fixed error message resulting from no result for functions that return no values

dbenn:
Hi @mpyat2. A small fix but nice to sort out.


mpyat2 (commented):
Thank you, @dbenn !

Issue #438, opened 2024-07-02
Generate more VeLa error messages 


dbenn:
For example, from `default` `switch` cases in `eval()` which may fail silently.

PR #439, opened 2024-07-09
#438: added common binary operator error method and made liberal use …

dbenn:
…of it; added more UTs


dbenn:
A bunch of error cases added in `VeLaInterpreter` @mpyat2

mpyat2 (approved):

PR #450, opened 2024-10-18
105 may need apply in vela

dbenn:
This is a bit oscure @mpyat2 so let me know if you have questions. It was really just a bug in function application in the presence of HOFs. The Y combinator stress tested it.

### DIFF
--- a/script/VeLa/Y/README.md
+++ b/script/VeLa/Y/README.md
@@ -1,0 +1,3 @@
+Steps on the way to the Y Combinator.
+
+Translated from JavaScript to VeLa.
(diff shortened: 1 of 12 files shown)


mpyat2:
Hmm... This is out of my knowledge, sorry. I completely trust you in all the VeLa things...

dbenn:
Thanks @mpyat2. Perhaps the simplest approach for future VeLa changes/fixes is to at-mention you in a comment so you are aware of a change that's coming, instead of asking you to review a pull request. The exception would be where a change to VeLa is likely to impact some other aspect of VStar functionality. Then I would ask for a review.

Issue #452, opened 2024-10-20
"Magnitude and Time Shifter" plug-in counterintuitive behaviour


mpyat2:
I'm not sure if it is a bug or a feature. 
1) Open observations in several different filters
2) Tools->Magnitude and Time Shifter: you will see the "Select a series" dialog
3) Click the "Deselect All" button. Not only will the checkboxes be cleared, but all the series will become invisible.
4) Click the "Select All" button. The checkboxes will be checked; all series will become visible.

On the other hand, clicking on checkboxes does not change the visibility of the series.

I'm also not sure the "Means" series makes sense here.

dbenn:
Thanks for reporting this @mpyat2. I understand what's causing the problem and will fix ASAP. I had generalised the dialog's components but missed a couple of checks.

It's a good question re: whether Means or other synthetic series, e.g. model and residual datapoints, make sense in the context of `ObservationTransformation` plugins. Initially I did not want to be prescriptive. I'm happy to remove Means, model, residuals, if you agree. Filtered series probably makes sense still since you can create series from such a series. Or perhaps only user-defined series created from a filter series should be included.

Let me know what you think.

mpyat2:
Hi @dbenn 
I agree with your approach. I think the filtered series can be included.

dbenn:
#453 created

PR #454, opened 2024-10-23
#452: added guards in select/deselect all button listener methods to …

dbenn:
…determine whether to modify visibility

Please check select/deselect bug fix @mpyat2


mpyat2:
Hi @dbenn, now the [Select all] [Deselect all] buttons do not affect the series visibility (as expected).

Shouldn't the Analysis group be removed from the dialog?
image

dbenn:
Try that @mpyat2. As per my comment on #452, I was initially thinking of this as a separate issue, but I think removing them as part of this issue is fine.

mpyat2:
Hi @dbenn ! Great! Thank you. It works.

Issue #459, opened 2024-12-14
DCDFT calculates slightly distorted power values.


mpyat2:
TSBase.statcomp() is called several times (at least twice) while calculating DCDFT. Since the 'dvar' variable (magnitude variation) is never initialized explicitly, the 'dvar' value used for the power estimation is slightly distorted (see TSBase.project()).

Since we are interested in relative power values, this error does not affect the end user.

To resolve the issue, it is sufficient to initialize 'dvar' in TSBase.statcomp().

dbenn:
Files to support Max's analysis: [files.zip](https://github.com/user-attachments/files/18138749/files.zip)

Thanks for the pick up and the analysis @mpyat2!

dbenn:
I looked at the original Fortran code and can see that `dvar` is not initialised locally in `statcomp` either:
~~~
c local variables
      integer n,nx
      real*8 dw,dt,dx,dtspan,xx,dts2

      call setlimit

      nbrake=0
      numact=0
      dweight=0.d0
      dave=0.d0
      dtave=0.d0
      dtvar=0.d0
      do n=nlolim,nuplim
       if (wvec(n).gt.0.d0) then
        numact=numact+1
        dw=wvec(n)
        dt=tvec(n)
        dx=xvec(n)
        dweight=dweight+dw
        dave=dave + (dw*dx)
~~~

PR #460, opened 2024-12-15
initialization of dtvar added in TSBase.statcomp()

mpyat2:
See #459


mpyat2:
Hi @dbenn 
I see errors in unit tests, inspecting them (FreqRangeTopHitsDcDftTest.java). Top hits were changed with the fix.

dbenn:
I'd expect some UT failures with this change.

mpyat2:
Failed:
1) CleanestTest
2) DcDftTest
3) FreqRangeTopHitsDcDftTest
4) StdScanTopHitsDcDftTest
I'll work on them.

mpyat2:
I updated the tests. Data for the tests were taken from Peranso output, except for the last two tests in CleanestTest.java:
testRefineOneFreqAndOneVariablePeriod()
and
testRefineOneFreqAndOneLockedPeriod()
For these two tests, I took data from the VStar output with the applied fix. It is not fair, but I could not find the exact analog in Peranso's CLEANest Workbench. I think it is not critical because, in each test, only power values change by a minor factor after the fix; the relative power values for all peaks are preserved.

dbenn:
Thanks @mpyat2 ! Looks good so far. Here's a case that shows that sometimes a small line-of-code count can have changes that impact more test code than product code. ;) Working through it.

dbenn (approved):
Thanks @mpyat2. Appreciate it! This one wins the prize for smallest code change leading to largest UT change. :)

PR #463, opened 2024-12-26
461 spectral window

mpyat2:
This implements the "Spectral Window" algorithm (with the results identical to the popular program Period04, http://period04.net/). The algorithm is essentially the 'classic' Discrete Fourier Transform as it is described in (Deeming 1975, bibcode 1975Ap&SS..36..137D) for a signal with unit amplitude.
Since we need DFT for this, I made the plug-in 2 in 1: DFT itself (again, identical to Period04, which uses this very algorithm) and Spectral Window. The 'Classic' DFT is the most primitive one and loses to other algorithms (for example, DC DFT) in many cases, yet why not have it for the collection?

To-do: 
1) Implement 'Create Model' functionality accessed directly from a panel with the DFT result (like in DC DFT). This button is hidden now, when I tried to use it 'as is', an error occured. It seems  'Create Model' is too tightly bound with DC DFT.
2) Unit test
3) Try to optimize calculations: now it is slower even as Peranso3 (where the same algorithm is implemented, too)

Side notes:
1) This plug-in, like DC DFT and AoV ones, may open several resulting non-modal dialogs. When a new star is loaded, they lose connections between their panes. So, I decided to close all of them while loading a new star. This is implemented with the help of List\<PeriodAnalysisDialog\>. It is probably worth implementing such behavior in DC DFT and AoV plug-ins.
2) Looks like there are still many places in VStar where Swing dialogs are called from threads other than EDT. This is strongly discouraged by the Java docs, strange, but it has no visible effect in the app. Nevertheless, I avoided it here with the help of SwingUtilities.invokeAndWait. Probably, this can be used in the other places in the app.

Test data: eta_Aql_2024_visual.txt 
[eta_Aql_2024_visual.txt](https://github.com/user-attachments/files/18254679/eta_Aql_2024_visual.txt)

image
image


dbenn:
Really cool @mpyat2! Perhaps the DFT will have issues with unevenly spaced data, which your comment may be alluding to, but this is a great addition! I am less familiar with Spectral Window, but see some references, e.g. https://encyclopediaofmath.org/wiki/Spectral_window If you have another reference you are using, could you point me to that?

As you may see from #204 I'm implementing a piecewise linear model (almost done) that will be incorporated with the AoV period search plugin for model creation as well as being standalone. Some of the other issues you see could be collaborated on if it helps.

mpyat2:
Hi @dbenn , concerning "Spectral Window": it is implemented like in Period04, see http://period04.net/files/p04manual_v1.0.0.pdf, page 53. There is no theory (unfortunately) in the guide, but only a practical example. I'll look for a better description. Also, a very lapidary description can be found here: https://www.cbabelgium.com/peranso/UserGuideHTML/Spectralwindow.html. I also wrote a couple of words about it in the documentation for the plug-in (see DFT and Spectral Window Plug-In.pdf)

mpyat2:
Side note about performance: on my mini-PC (low-performance machine, Celeron N5095A @ 2.00GHz, Windows 11) DFT calculations for the test data "eta_Aql_2024_visual.txt" (minFreq=0.0, maxFreq=2.0, resolution=0.00001) with Java 8 (1.8.0_432, last update) takes ~41s, while with Open JDK 23.0.1 it takes only ~7s (almost in 6 times faster!).

Well, recent versions of Java have done some great optimization...
image

dbenn:
Recommending a new minimal version of Java of VStar may be worth considering @mpyat2.

mpyat2:
> Recommending a new minimal version of Java of VStar may be worth considering @mpyat2.

Well, it's probably sufficient to indicate that it it recommended to switch to the newest version available because of better optimization, performance win, etc.

mpyat2:
> I am less familiar with Spectral Window,

Also, the Spectral Window can be considered as a 'representation of gaps' and the finite nature of the observing interval. The Fourier spectrum of the real data is a convolution of an 'ideal' spectrum with the Fourier transform of this "gaps' function" (see this abstract https://ui.adsabs.harvard.edu/abs/1975Ap%26SS..36..137D/abstract). By the way, in the appendix to the article, a short Fortran program, which calculates both DFT and Spectral Window, is listed. Well, I noticed it now only; in fact, it is quite simple and is easy to reinvent (this is a straightforward calculation of Eq. 3 from the article).
In other words, if you calculate the spectral window (which is centered at 0), you will see all phantom structures that accompany the real spectral peaks.

dbenn:
> > Recommending a new minimal version of Java of VStar may be worth considering @mpyat2.
> 
> Well, it's probably sufficient to indicate that it it recommended to switch to the newest version available because of better optimization, performance win, etc.

I have found issues with some of the newer versions with respect to Swing drop-down list collection widgets. Other than that, agreed. Some testing required.

mpyat2:
Hi David, @dbenn 
I implemented multiPeriodicFit for this DFT using org.apache.commons.math.stat.regression.OLSMultipleLinearRegression.
I still need more tests yet it (seems it) works identically to the current implementation of it for DC DFT / Fourier Model.
OLSMultipleLinearRegression is quite flexible, and I plan to make an extended Fourier Model plugin based on it with [algebraic] polynomial trend approximation in addition to the existing trigonometric polynomial fitting. So it will be Fourier Model + Polynomial Fit "2 in 1", but it requires additional work and I cannot promise anything now :)

dbenn:
Very cool @mpyat2. A bit slow to get to the review and trying this. Apologies. Will get to it in the next day or so. I will also have something for you to review for me soon. :)

mpyat2:
No problem, David, thank you! Anyway, I still need to create a unit test for it.

mpyat2:
Hi David @dbenn ,
I've added multithreading to this plug-in and DC DFT based on OLSMultipleLinearRegression. It gives the same result as Foster's code; however, it is noticeably slower. Nevertheless, with multithreading, the calculations are faster starting from a 4-core CPU (tested under Debian (very old Dell desktop) and Windows (relatively new yet cheap MiniPC with a 4-core Celeron)). On a 12-core I7 Windows laptop, the multithreaded OLSMultipleLinearRegression-based variant is several times (~3 times) faster.
So, the plug-in combines DFT (Deeming), Spectral Window (based on Deeming's DFT), and Ferraz-Mello's DC DFT.
image

mpyat2:
Hi @dbenn 
I made small changes in the plug-in, and something got broken in the unit tests for Java 9+... Could you please look at them? I cannot figure out what is wrong.

dbenn:
> Hi @dbenn
> I made small changes in the plug-in, and something got broken in the unit tests for Java 9+... Could you please look at them? I cannot figure out what is wrong.

I recently had some non Java 8 test failures for the plugin I'm working on as well @mpyat2. Odd. Will see what I can find unless you beat me to it :).

mpyat2:
Hi @dbenn 

I've done local tests with
* java-se-8u44-ri
* openlogic-openjdk-8u432-b06-windows-x64
* jdk-9.0.4
* jdk-20.0.1

Plugin tests succeeded in all cases.
VStar tests failed. In all cases, VSXWebService-related functions failed; PluginManagerTest.testRetrievePluginInfo() failed for java-se-8u44-ri. 

See the attached file:
[_tests.zip](https://github.com/user-attachments/files/18438705/_tests.zip)

Also, I needed to set JAVA_TOOL_OPTIONS=-Duser.country=EN -Duser.language=en before running the tests. In my locale, a comma is used as a decimal separator, which causes problems in some tests where floating-point numbers are compared.

dbenn (approved):
Thanks @mpyat2 !

PR #464, opened 2025-01-25
204 piecewise linear model plugin from current mean series

dbenn:
Hi @mpyat2. There are 3 main additions here:

1. A piecewise linear model plugin.
2. A change to the AoV period search plugin to be able to create a piecewise linear model corresponding to the phased binned means of the selected top hit.
3. A plugin library that implements most of this, and which is used by 1. and 2.

The created model includes fit metrics and extrema and there are regular JUnit tests for the library.

I have refactored quite a lot of the modelling code although I think merge conflicts are confined to the plugin build.xml.

Note that the VeLa script commits are unrelated to the model so feel free to ignore.


mpyat2:
1) Open a data file (see attached one as an example)
[V965_Cep_PMAK_BJD_simple.tsv.zip](https://github.com/user-attachments/files/18549533/V965_Cep_PMAK_BJD_simple.tsv.zip)
2) AoV with period range -> Min Period = 0.08, Max Period = 0.09, Resolution= 0.0000001, Bins=10
3) Select a top hit: P= 0.0850673, click [Create Model]
**Error:**
image
4) Click the [New Phase Plot] button
5) Click [Create Model] again
image
6) Then return to the 'Raw data' view and zoom
image
7) **Where are the model points?**
image

dbenn:
Investigating.

dbenn:
After a distracted week, I have looked at the disappearing model data points again.

This error does not occur when using the PLF model creation plugin in isolation from AoV plugin.

dbenn:
The model data points that are not visible within a **_particular isolated segment_** of the plot at a **_particular zoom level_** appear to be deterministic, at least.

dbenn:
Okay, so the zoom happens in two parts: zoom range and domain.

I have determined that it is the **domain zoom** that causes the problem, since disabling this:

https://github.com/AAVSO/VStar/blob/master/src/org/aavso/tools/vstar/ui/pane/plot/AbstractObservationPlotPane.java#L629

such that when only range zoom happens, the data points do not disappear.

dbenn:
Moving the domain and range zoom lines makes no difference to the outcome.

dbenn:
Trying newer JFreeChart version next.

dbenn:
So, using the latest version of JFreeChart (1.5.5 vs 1.5.2) makes no difference.

dbenn:
I also tried creating the model, copying the VeLa function from the Models dialog, loading the observation file again, creating a phase plot at the period we found already (from AoV period search, for example), then using the VeLa Model plug-in to create a model.

This counts as creating the same model in two different ways.

The result is the same. Zooming into the phase plot with model datapoints present is fine. When switching to raw mode from phase plot mode, the same zoom error (when x axis zoom happens in the code) occurs.

dbenn:
See my comments above @mpyat2.

This one is very puzzling.

I'm tempted to release the plug-in with a documented Known Bug since creating the model in phase plots is useful by itself.

But... I'd prefer to resolve this problem, so will keep trying for now. Thoughts welcome.

dbenn:
@mpyat2 I'm **_close_** to the point at which I am considering documenting this as a known bug, making this plugin available in the next release, and moving on. The bug should be fixed, of course, but I think the plugin and its use in AoV have enough utility to make what exists available. What's your view?

mpyat2:
> @mpyat2 I'm **_close_** to the point at which I am considering documenting this as a known bug<...>

Hi @dbenn , I agree at this point. Let's describe it as a known bug.

dbenn:
I'll merge if you are happy @mpyat2 but if you have any code comments that's fine of course.

mpyat2:
Hi @dbenn , I do not have comments at this point.

mpyat2 (approved):

PR #473, opened 2025-03-05
472 fix various test failures

dbenn:
Hi @mpyat2, okay, all UTs are passing and we have Java versions 8, 11, 17, 21, 23.


mpyat2 (approved):
Everything looks good

PR #476, opened 2025-03-08
Create pmd.yml

dbenn:


dbenn:
@mpyat2 I'm adding security static analysis checks.

dbenn:
Backing out of this for now due to errors. Will revisit static analysis later.

Issue #456, opened 2024-11-12
Should be able to replace function with lambda in parameter type


dbenn:
It should be possible to use the lambda symbol where `function` is now used in a function parameter, e.g.
~~~
f(g:λ) {
...
}
~~~
instead of:
~~~
f(g:function) {
...
}
~~~
Noticed this after looking at Y in #105.

dbenn:
Example:
~~~
f(g:function x:integer):integer {
 g(x)
}
~~~
vs
~~~
f(g:λ x:integer):integer {
 g(x)
}
~~~
with example invocation:
~~~
f(λ(x:integer):integer{x*x} 4)
~~~

PR #477, opened 2025-03-10
#456: added lambda as an alternative to function as a type, with the …

dbenn:
…same meaning; added UTs for this and updated example Y.vela


dbenn:
Up to you re: whether you want to review this @mpyat2.

Issue #480, opened 2025-03-27
Chart font tuning via 'preferences'


mpyat2:
Creating additional chart-related settings that allow customization of fonts.

mpyat2:
Like this:

![Image](https://github.com/user-attachments/assets/47b39ec0-fa93-47ca-ac7b-1825aba561d4)

PR #481, opened 2025-03-27
Customizable chart fonts

mpyat2:
Hi David, 

I've added the customization of chart fonts (see #480). I wanted to make the chart more 'sexy' for publications; well, the quality of PNGs is still not perfect, but could become slightly better after the manual font tuning (IMHO).

What do you think?

(I probably do additional checks and tunings)


dbenn (approved):
Hi @mpyat2. This is great! I've tested it and all looks good. Thanks!

PR #483, opened 2025-03-30
The chart's right margin ('padding') is set to a non-zero value

mpyat2:
The chart's right margin ('padding') is set to a non-zero value to avoid clipping the rightmost X-label.


dbenn (approved):
Thanks @mpyat2. This is something I have meant to fix for a long time.

PR #491, opened 2025-05-01
Error selecting the work dataset in DFTandSpectralWindow fix

mpyat2:
Now the FtResult object is always recreated with the selected dataset as a parameter.


dbenn (approved):
Thanks @mpyat2

Issue #495, opened 2025-05-18
DFTandSpectralWindow plug-in test fails on Mac with openjdk 17.0.5 (numeric tolerance)


dbenn:
On my Mac (13.4) with openjdk 17.0.5,  `DFTandSpectralWindow`'s plug-in test fails:
~~~
TopHits Power comparison failed: expected 103.32745790177243; got 103.32745788221; obs.#1
TopHits Power comparison failed: expected 103.32745790177243; got 103.32745788221; obs.#1
[FAILED]  DFTandSpectralWindow ('DFT and Spectral Window Frequency Range') test
** PASSED: 10, FAILED: 1, NO TEST: 42
~~~
Changing:
~~~
compareDftResult(expectedDftTopHits, topHits, "TopHits ", 1e-8);
~~~
to:
~~~
compareDftResult(expectedDftTopHits, topHits, "TopHits ", 1e-7);
~~~
in `dftFerrazMelloTest()` resolves this.

This error is not happening in GitHub Actions, so may be a floating point difference due to Apple Silicon vs the Action runner's platform.

dbenn:
@mpyat2: are you happy for this small change to be made? In anticipation, I'll created a branch and pull request. Thanks.

PR #496, opened 2025-05-18
#495: changed numeric tolerance in compareDftResult() called by dftFerrazMelloTest() from 1e-8 to 1e-7

dbenn:
Let me know what you think @mpyat2. CI/CD is happy too.


mpyat2:
Hi @dbenn , this is fine. I experienced a similar problem with tests for the "DFT and Spectral Window" plug-in.

mpyat2 (approved):
It's OK

PR #499, opened 2025-05-19
Fix of the calcPopVariance() method in the DFTandSpectralWindow plug-in.

mpyat2:
The previous calcPopVariance() implementation sometimes produced imprecise results. The current is more robust.


dbenn (approved):
Thanks @mpyat2

PR #510, opened 2025-06-05
Added progress indicator to the DFTandSpectralWindow plug-in like in AbstractObservationRetriever

mpyat2:
1) The progress indicator has been added
2) In AbstractObservationRetriever: public void incrementProgress(int steps) has a bug (steps is ignored). This method is currently never used. I had commented it out, however, it can be easily fixed:
the second parameter in "new ProgressInfo(ProgressType.INCREMENT_PROGRESS, 2)" should be replaced with _steps_.

Such a progress indicator might be added to the other PeriodAnalysisPluginBase descendants (the "incrementProgress" method might be included in PeriodAnalysisPluginBase )


dbenn (approved):
Thanks @mpyat2

Issue #532, opened 2025-06-23
Allow shortenings of some VeLa types, keywords


dbenn:
Types:
- `ℤ` or `int` for integer
- `ℝ` for `real`
- `bool` for `boolean`
- `str` for `string`
- `fun` for `function` (we already allow `𝛌`)

VeLa dialog choosers / key shortcuts for `ℤ`, `ℝ`, `𝛌`, `π` should also be added.

dbenn:
Example usage in functions:
~~~
f(n : ℤ) : ℤ {n*n}
f(n : int) : int {n*n}
f(n : integer) : integer {n*n}
~~~
~~~
f(n : ℝ) : ℝ {n*n}
f(n : real) : real {n*n}
~~~
~~~
f(b : bool) : bool {not b}
~~~

dbenn:
Add to filter obs dialog text area as well.

dbenn:
I initially thought that `𝔹` was not a suitable symbol for the set of Boolean values, but I've since seen at least one example of usage: in the language _Agda_. I also initially had problems getting it to render via Unicode values in Java. I have that sorted out now.

PR #534, opened 2025-07-07
#532: added shortened/alt names for types

dbenn:
Hi @mpyat2. I've added shortcuts for some types as per #532 including escape-based keyboard shortcuts to the VeLa dialog, e.g. type `\` (which won't display) followed by `l`, `p`, `r` or `z` to see the effect (the `\` is consumed so you won't see it). More such shortcuts may be added over time. I removed `\b` for Boolean for now because of some contention over what Unicode character could/should be used.

### DIFF
--- a/src/org/aavso/tools/vstar/plugin/filter/impl/VeLaObservationFilterDialog.java
+++ b/src/org/aavso/tools/vstar/plugin/filter/impl/VeLaObservationFilterDialog.java
@@ -149,7 +149,8 @@ public class VeLaObservationFilterDialog extends AbstractOkCancelDialog {
 			@Override
 			public void actionPerformed(ActionEvent e) {
 				String propText = (String) obsPropsList.getSelectedItem();
-				velaFilterField.append(propText);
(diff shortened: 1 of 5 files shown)


mpyat2:
Hi @dbenn, it works for me. However, what if someone wants to enter a backslash (\\) as part of a string? An additional escape sequence (\\\\ -- double backslash) could be a solution. What do you think? 

image

dbenn:
> Hi @dbenn, it works for me. However, what if someone wants to enter a backslash (\) as part of a string or a comment? An additional escape sequence (\\ -- double backslash) could be a solution. What do you think?

Agreed  @mpyat2. I thought about this then forgot. Thanks for the reminder.

dbenn:
Have added the double backslash but doesn't seem to have the desired effect.

mpyat2:
Thank you @dbenn , it works
image

mpyat2 (approved):

dbenn:
Thanks @mpyat2

Issue #535, opened 2025-07-17
Update supported bands in SeriesType and AID dialog


dbenn:
An AAVSO member contacted Bert Pablo to ask whether U was supported in VStar.

This is the first check/update of `SeriesType` in awhile.

dbenn:
Ran `python ./script/gen_bands_enums_from_ws.py` and as it turns out, Johnson U is already supported, but I noticed that Clear Blue Blocking is not listed now and that Gaia G has been added.

Can leave Clear Blue Blocking for legacy's sake and will add Gaia G.

dbenn:
@mpyat2, in conversation with @bpablo-aavso, Johnson U was missing from the AID pane, so I have added it. It was already in `SeriesType`.

I have also added to the AID dialog a "Spacecraft" tabbed pane and added Gaia G to it. Not sure that's the best tab name so, @mpyat2 and @bpablo-aavso, let me know if you have thoughts about that.

dbenn:
There's formatting changes in some of the code. Hopefully not too distracting.

PR #536, opened 2025-07-18
#535: added Gaia G

dbenn:
Hi @mpyat2, please review this small change

### DIFF
--- a/src/org/aavso/tools/vstar/data/SeriesType.java
+++ b/src/org/aavso/tools/vstar/data/SeriesType.java
@@ -37,808 +37,733 @@ import org.aavso.tools.vstar.util.notification.Notifier;
  */
 public class SeriesType implements Comparable<SeriesType> {
 
-	// Static members
+    // Static members
 
-	private final static int NO_INDEX = -1;
+    private final static int NO_INDEX = -1;
 
-	// Plot point size.
-	public final static int DEFAULT_SIZE = 4;
+    // Plot point size.
+    public final static int DEFAULT_SIZE = 4;
 
-	private final static String COLOR_PREFS_PREFIX = "SERIES_COLOR_";
-	private final static String SIZE_PREFS_PREFIX = "SERIES_SIZE_";
+    private final static String COLOR_PREFS_PREFIX = "SERIES_COLOR_";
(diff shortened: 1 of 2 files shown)


mpyat2:
Hi @dbenn ,
In the Gaia DR2/DR3 observation source plug-in, the Gaia G magnitude is defined somewhat differently, along with the Gaia BP and RP magnitudes. I propose adding the definitions for Gaia BP and RP to SeriesType.java, which would then allow me to remove the Gaia band definitions from the plug-in. Thought?
image

dbenn:
Hi @mpyat2. That makes sense to me. If there's one Gaia band in `SeriesType`, they probably all should be. I'm only hesitating because:
- I'm trying to understand why AAVSO are only officially supporting Gaia G right now vs the other two, e.g. is there something about their status that's different?
- I've emailed @bpablo-aavso to ask about this too. Bert you may wish to respond in this issue instead.
- We're mixing code that would normally only be in a plugin vs core VStar code.
But on balance, I like your suggestion and it seems fair.

bpablo:
Hello David,

When we talk about bands that we support, this tends to be only bands that could actually be taken by a ground-based telescope and put into the AID by an individual. The GAIA bandpasses are all custom and are only used by GAIA. When we added GAIA G it was at the request of someone using a space based telescope who was transforming to GAIA G because it was a much closer match to the actual bandpass than any of the Johnson or Sloan filters. It is highly unlikely this will ever be the case for the other GAIA filters, which is why we haven't added them. Does this make sense?

dbenn:
That does make sense @bpablo-aavso, yes. Thanks.

@mpyat2, are you satisfied with Bert's explanation? It seems that we're saying that the other two bands should stay in the plugin code.

My spacecraft tab may not be necessary if there's only one band. @bpablo-aavso, would the Color tab make sense for this?

mpyat2:
Hi @bpablo ,

I'm slightly confused. AID is a database that mostly collects observations carried out by amateurs. How can they use the Gaia G band?

mpyat2 (approved):

mpyat2:
Hi @dbenn 
I'll remove Gaia G from the plug-in at the next step.

Issue #88, opened 2020-11-22
JD to HJD converter plugin is consistently different from other tools by a small amount


dbenn:
There is a lot of detail in https://sourceforge.net/p/vstar/bugs-and-features/680/ so we'll keep that open until the issue is addressed.

dbenn:
It's worth noting that the current implementation is based upon Meeus's low accuracy method. The high accuracy method also exists.

dbenn:
Having this assistance is valuable. I've been a skeptic and have had (and still do) concerns about what GenAI is doing to us, the environmental impacts, and the importance of verification and the potential for many to ignore the need for it and blindly accept what GenAI gives it. This activity is one sweet spot though. There are also I think likely to be others that will improve both productivity and creativity.

However, we're not done. The GenAI claimed that the results were better. In some cases they were, and in other cases not. In particular, while the first 2 `testHJD*()` method results were improved, the last 4 were not. My next conversation in Cursor suggested a few plausible reasons relating to:

- solar coordinates
- radius vector
- position dependence
- low vs high accuracy solar model (which has been in the back of my mind since I wrote the code)

I am going to look at each. What my exchanges so far on this topic illustrate is that it is important to not blindly trust the AI, but to ask: does this make sense? It does feel a bit like a superpower though, I have to admit.

dbenn:
Full transcript of interaction with Cursor leading to the eccentricity fix: [cursor_initial_HJD_fix_eccentricity.md](https://github.com/user-attachments/files/25314758/cursor_initial_HJD_fix_eccentricity.md)

It started with me seeing what Cursor could summarise and do.

dbenn:
I changed to code (instead of asking Cursor to create patch files this time) to use J2000 solar longitude and the results improved further. Way further. The worst difference between VStar's HJD conversion code and BAA's is in the order of 3 hundredths of a second. Some results are closer to 3 thousandths of a second delta!

dbenn:
Full transcript of interaction with Cursor leading to the eccentricity and J2000 fixes: 

[cursor_transcript.md](https://github.com/user-attachments/files/25315463/cursor_transcript.md)

It started with me seeing what Cursor could summarise and do.

dbenn:
[...]

PR #557, opened 2026-02-18
88 jd to hjd converter plugin is consistently different from other tools by a small amount

dbenn:
@mpyat2, I have introduced changes to fix this long-standing JD to HJD conversion problem.

@clkotnik, I wanted to include you as a reviewer here (even if only for your awareness) since we had interactions about this bug including a forum post you wrote, an embarrassingly long time ago. I got there in the end. I probably should have more strongly considered your suggestion at the time of using alternative Java code.

In the end, my robot overlords (Cursor) helped me. This use of GenAI and agentic AI in particular is a sweet spot I think. Please see  issue #88 for commentary.

### DIFF
--- a/script/jd2hjd.py
+++ b/script/jd2hjd.py
@@ -1,0 +1,30 @@
+from astropy.time import Time
+from astropy.coordinates import SkyCoord, EarthLocation
+import astropy.units as u
+
+import sys
+
+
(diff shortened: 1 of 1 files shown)


dbenn:
> Hi @dbenn , please give me a couple of days. I’ll look at it over the weekend.

Thanks @mpyat2 :)

PR #559, opened 2026-03-09
558 wwz statistic goes to zero in the presence of gaps

dbenn:
Hi Max. The fix for this long-standing bug also benefited from a collaboration with Cursor. Please see what you think.


mpyat2 (approved):
Hi @dbenn , it seems the solution works. Thank you!

PR #610, opened 2026-06-05
#605: make VeLa numeric literals locale-independent

### LINKED ISSUE #605: Portability of VeLa code with localised numbers
The same code will work in one locale but fail in another, e.g.
~~~
x is -4,2
~~~
vs
~~~
x is -4.2
~~~
The idea originally was to allow flexibility via localised numbers.

While continuing to allow this, should we also always translate localised decimal point characters such as `,` to `.` internally and apply numeric parsing via the same locale?

See https://forums.aavso.org/t/flux-to-magnitudes/4178/11

dbenn:
## Summary

Fixes #605. VeLa accepts both `.` and `,` as the decimal point, but `parseDouble()` interpreted the literal through `Locale.getDefault()`, so the *same* code worked in one locale and failed in another (e.g. `-4.2` vs `-4,2`).

A comma can only ever appear **inside** a `REAL` token as the decimal point: lists and call arguments are whitespace-separated, comma's only grammatical role is the `POINT` fragment (`POINT: PERIOD | COMMA`), and a `REAL` admits a single separator. So we can normalise `,` → `.` losslessly and parse in a locale-independent way, keeping the original flexibility of localised numbers while making the same source behave identically everywhere.

## Changes

- **`ExpressionVisitor.parseDouble`**: normalise `,` → `.` and parse with `Double.parseDouble`. Removes the locale-dependent `DecimalFormat` / `DecimalFormatSymbols` / `ParsePosition` machinery and the `setExponentSeparator` / `setMinusSign` workarounds (`Double.parseDouble` already handles `e`/`E`, a leading `.`, etc.).
- **`VeLaTest.commonNumericLocaleTest`**: fix a latent bug where the `prog` parameter was reassigned inside the locale loop (stripping all `.` after the first non-dot locale, so later locales tested an already-converted string); restore the default locale in a `finally`; and only substitute decimal separators VeLa actually supports (`.`/`,`), otherwise verify a `.`-written program still parses under that default locale.
- **New portability tests**: `testDecimalPointPortableAcrossLocales`, `testDecimalCommaPortableAcrossLocales`, `testExponentPortableAcrossLocales` (plus a `commonNumericPortabilityTest` helper) confirm that `.` and `,` literals evaluate identically across all available locales without rewriting the source.

## Notes

- This keeps `.` and `,` as the only supported decimal points (by grammar design). Locales whose separator is some other character (e.g. Arabic `٫`, U+066B) aren't lexable as numbers and are out of scope here — handling them would require a grammar change.
- Relies on `,` keeping its current role (decimal point only). If a future change gives comma a syntactic meaning (e.g. an optional list/argument separator), that decision should come first.

## Test plan

- [x] `ant test` / `VeLaTest` — all 222 tests pass
- [x] Verified under both the default (CLDR) and `COMPAT,CLDR` locale providers
- [x] `4.2` and `4,2` both evaluate to `4.2` regardless of default locale

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -1,16 +1,11 @@
-# This workflow will build a Java project with Ant and create a daily 
-# release of the bash and win targets.
+# This workflow will build a Java project with Ant, create a daily release of the 
+# bash and win targets, update the grammar railroad diagram navigable from the wiki.
 #
 # For more information see:
 # - https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-ant
 # - https://github.com/AAVSO/VStar/issues/340
 
 name: Snapshot release 
-
-#on:
-#  schedule:
-#     # every day at 0845 GMT
-#    - cron: '45 08 * * *'
 
 on:
   push:
@@ -44,9 +39,6 @@ jobs:
     - name: Build win archive
       run:
         ant -noinput -buildfile build.xml win
-#    - name: Build mac archive
-#      run:
-#        ant -noinput -buildfile build.xml mac
     - name: Build plugins archive
       run: |
         cd plugin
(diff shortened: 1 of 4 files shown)


mpyat2 (approved):
Thank you, David! It works! Among other benefits, it resolves a problem with #VELAFILTER directive in flex-format files: now it can be interpreted locale-independent.

PR #614, opened 2026-06-18
611 create a plug in that implements mmrd method of nova distance determination

### LINKED ISSUE #611: Create a plug-in that implements MMRD method of nova distance determination
Two papers relevant to nova distance determination with the MMRD method are:

- https://arxiv.org/pdf/1703.04087
- https://www.aavso.org/sites/default/files/jaavso/v38n2/193.pdf

What we want is to write a plug-in that does for novae what the Leavitt's Law distance calculator plug-in does for Cepheids.

dbenn:
Please have a look over this MMRD nova distance calculator @mpyat2


mpyat2 (approved):
Hi @dbenn, this is a very interesting tool.

PR #619, opened 2026-07-25
#608: add preferences for default source and sink directories.

### LINKED ISSUE #608: Add a preference setting for default data directory
See AAVSO forum message: https://forums.aavso.org/t/vstar-2-25-0-set-default-directory/4510

DEY_VAR asks:
> Is there a way to set the default startup directory when running a local Vstar?
> 
> My local observation files don’t live in the windows top of the file system that Vstar starts up in. Many layers below and not on the system disk so the user has many clicks to get into the general area of the file system where my data files live.
> 
> Thanks!

This could be individualised per observation source plugin but I suspect a single path is enough.

We should add a Preference pane for this and other default paths, e.g. output file path.

dbenn:
This change has been requested recently and a few times in the past @mpyat2.

It's just a new Preferences pane and changed obs src and sink directory for file choosers as a result.

Feel free to try it out. If you don't have time, that's OK. It's a fairly limited change.

### DIFF
--- a/src/org/aavso/tools/vstar/ui/dialog/AdditiveLoadFileOrUrlChooser.java
+++ b/src/org/aavso/tools/vstar/ui/dialog/AdditiveLoadFileOrUrlChooser.java
@@ -43,6 +43,7 @@ import org.aavso.tools.vstar.ui.resources.PluginLoader;
 import org.aavso.tools.vstar.util.Pair;
 import org.aavso.tools.vstar.util.help.Help;
 import org.aavso.tools.vstar.util.locale.LocaleProps;
+import org.aavso.tools.vstar.util.prefs.FilePathPrefs;
 
 /**
  * This class aggregates a JFileChooser and additive load checkbox and URL entry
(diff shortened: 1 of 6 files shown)


mpyat2:
Hi @dbenn , the preferences work.
One comment.
Now the open/save dialogs (file choosers) always open with these directories as the initial ones, even if the user changed the directory in a file chooser. Previously, when the default directory was "Documents" (in Windows), after changing the directory while opening/saving files, the choosers opened in that changed directory. Now they always start from the configured directories.
Would it be better if only the first invocation of the open/save dialogs used the configured directories as starting points?

dbenn:
Apologies @mpyat2. I thought I had tested this but apparently had not. Fixed. Please try again.

mpyat2:
Thank you, @dbenn! It's perfect now. The thing I've wished for for a long time.

mpyat2 (approved):

dbenn:
Thanks @mpyat2 !

PR #622, opened 2026-08-22
Plugin's preferences via IPlugin.getPreferencesPane().

### LINKED ISSUE #334: Allow plugins to offer a Preferences pane that they know how to make use of
These could be added as tabs per plugin in the Preferences Plug-ins tab.

mpyat2:
Issue #334. BJDConverter.java, JDtoBJDTool.java preferences (common preferences pane)

### DIFF
--- a/plugin/src/org/aavso/tools/vstar/example/plugin/CustomFilterTest.java
+++ b/plugin/src/org/aavso/tools/vstar/example/plugin/CustomFilterTest.java
@@ -1,5 +1,6 @@
 package org.aavso.tools.vstar.example.plugin;
 
+import java.awt.Component;
 import java.util.List;
 
 import org.aavso.tools.vstar.data.SeriesType;
@@ -53,5 +54,13 @@ public class CustomFilterTest extends CustomFilterPluginBase {
 	@Override
 	public String getDisplayName() {
 		return "Previous Observation Condition";
+	}
+	
+	/**
+	 * @see org.aavso.tools.vstar.plugin.IPlugin#getPreferencesPane()
+	 */
+	@Override
+	public Component getPreferencesPane() {
+		return null;
 	}
 }
--- a/plugin/src/org/aavso/tools/vstar/example/plugin/ObservationSink.java
+++ b/plugin/src/org/aavso/tools/vstar/example/plugin/ObservationSink.java
@@ -17,6 +17,7 @@
  */
 package org.aavso.tools.vstar.example.plugin;
 
+import java.awt.Component;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.util.List;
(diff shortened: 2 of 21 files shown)


dbenn (commented):
This is excellent. Thanks @mpyat2! The only thing I'm wondering is whether plugin prefs panes could instead be part of tabbed panes in the Plugin Prefs tab to prevent an endless number of tabs at the top level. We could go with what's there for now and refine down the track if it becomes necessary. Thoughts?

mpyat2:
Hi @dbenn , great idea -- nested panes. I'll try do do this

mpyat2:
One non-obvious thing: when running from Eclipse, the Preferences panel for BJDConverter and JDtoBJDTool is common (by design), but stand-alone VStar shows two panels. Apparently, this is related to the way jar files are loaded. Will think about it. One possible solution: check unique id or something like this instead of comparing instances directly.

dbenn:
Hmm. That's interesting @mpyat2. Let you know if I think of anything. I can replicate what you describe though.

dbenn:
I have a possible fix @mpyat2 (just one tab in both Eclipse and ant build and run) along the lines of what you were thinking. Want me to commit that change to the branch for you look at or just add the diff here? Some assistance from Cursor on this one.

mpyat2:
Hi @dbenn , yes, of course!

dbenn:
Let me know whether this latest commit does what you expect. Looks OK to me.

mpyat2:
Hi @dbenn, I made changes to BJDConverter.java and JDtoBJDTool.java that should solve the problem with separate copies of ConverHelper. I also slightly changed PreferencedDialog.java: I removed the local List<Component> addedPanes, since it actually duplicates the existing pluginPrefs.

There's also some redundancy in PreferenceTabDeduper.recordPreferenceTab(): pane.getName() returns the prefsId that was just added right before it, so addPreferenceId(pane.getName(), seenIds) does nothing. What do you think?

I have not resolved the 'nested panes' improvement yet, so this is not the final commit

dbenn (commented):
This is looking great @mpyat2, thanks! The only thing I'm wondering is whether the old `Plug-ins` prefs tab and new `Plug-ins Preferences` tab will cause any confusion. Is it better to have the old `Plug-ins` tab have a sub-tab such as `General`, then with the per plug-in tabs that now appear in `Plug-ins Preferences`?

mpyat2:
> This is looking great @mpyat2, thanks! The only thing I'm wondering is whether the old `Plug-ins` prefs tab and new `Plug-ins Preferences` tab will cause any confusion. Is it better to have the old `Plug-ins` tab have a sub-tab such as `General`, then with the per plug-in tabs that now appear in `Plug-ins Preferences`?

Hi @dbenn , great idea! I'll try to do it

mpyat2:
Hi @dbenn, it seems that's all. I got rid of the separate "Plug-in preferences" pane,

dbenn:
Excellent! Thanks @mpyat2

dbenn (approved):

PR #623, opened 2026-08-26
#601: fixed JavaScript handling for Java 15+

### LINKED ISSUE #601: Tool -> Run Script... fails
The change of minimum Java version perhaps?

dbenn:
Running JavaScript seems to work again now, e.g. script/JavaScript/example.js

Let me know if you see any issues @mpyat2


mpyat2:
Hi @dbenn , in general, it works (example.js).
adopt.js works when I set existing directory, if the directory does not exist, it silently fails (no error messages).

dbenn:
@mpyat2 Yes that's a very good point. Should be fixed although it's separate from the bug since it's just something the script should do, e.g. open a dialog (`alert` doesn't work in this context), write to a console (which there is none at the moment). There's a `getError()` function available to JavaScript code.

mpyat2 (approved):

dbenn:
Hmm. I have taken a closer look at the reason for the save failure you saw @mpyat2. Please see the last commit for this. At least there will be a log entry now, e.g.
~~~
Sept 12, 2026 1:22:56 PM org.aavso.tools.vstar.scripting.ScriptRunner runScript
SEVERE: Cannot save plot to '/Users/david/Desktop/adopt-a-star/Landolt-V0348 Sgr_plot.png
~~~
More could be done and the way errors get reported depends on whether we're in scripting mode.

mpyat2:
Hi @dbenn , yes, now I can see the error in log:
Sep 14, 2026 4:47:07 PM org.aavso.tools.vstar.ui.dialog.MessageBox showErrorDialog
SEVERE: Cannot save plot to '\Users\david\Desktop\adopt-a-star\Landolt-V0348 Sgr_plot.png'.
Thank you for the improvement!

PR #627, opened 2026-09-09
#611: compare MMRD relations without repeating the exponential fit.

### LINKED ISSUE #611: Create a plug-in that implements MMRD method of nova distance determination

dbenn:
Keep t2/t3 after the first fit, show a sortable summary table of every calibration, and return to the input dialog to try another relation.

Basically allow the user to retry different methods and get a summary of all methods' results.


mpyat2:
Hi @dbenn, very interesting plug-in! It works. I tested it on the V4633 Sgr visual observations from the AID using 2450894.5 ≤ JD ≤ 2450954.5, and it gave me a distance of ~13 kpc (with no extinction, well, not far from your estimation). It seems you used different data? It is not entirely clear from the existing MMRD pdf.

mpyat2 (approved):

dbenn:
It could be either data or the method selected.

mpyat2:
> It could be either data or the method selected.

Sure, it depends on the data; however, it is not clear from MMRD.pdf what data was used (or did I miss the reference?)

dbenn:
You are 100% right @mpyat2 !

Sorry. the figures I've added there in the latest update for the doc don't correspond to V4633 Sgr it seems. I mention that target in Kok (2010) related sections but what's shown in the dialogs is something else.

There are other things that need to be cleaned up in the doc as well.

I'll create a new branch to fix the doc.

Thanks for picking this up!

